Issue #14, opened 2021-08-22
Undrop Units


adamisntdead:
Should be able to add back units that were previously dropped

(I will fix this myself and submit a PR)

PR #15, opened 2021-08-24
Allow a student to 'undrop' units

adamisntdead:
## Why

Previously, if a student had a unit they could then drop it and loose access. If this is done by accident or a student changes their mind, they shouldn't have to re-request (considering it had already been approved), as long as they are under the unit limit.

Addresses: #14 

## How

* Adds a new auto-approval condition that addresses the 'unit previously dropped' case


vEnhance:
This is nice, thanks
The tests should pass once it gets merged onto main since main uses pyflakes and not flake8

Issue #37, opened 2021-08-31
Clearer error message when required PDF missing


vEnhance:
Right now when a sufficiently large PDF is uploaded (more than 10MB or so) there
is a limitation on my systems where the file does not make it to the server.
From the server's perspective, the form was submitted with no file at all (even
though the client's browser should have stopped them in such a situation).

It would be good to have a clearer error message in this case on the submission
form stating that the user may want to check the file size as a possible
culprit.

PR #58, opened 2021-09-29
Solve (?) Issue #37 .

ETS1331:
The code here just alerts you if there is a problem with your content, which happens to be when it isn't a possible file submission (>10 MB or wrong file type); however, it doesn't detect when it is a wrong file submission or if there is no submission at all (thus the (?)).


vEnhance:
Thanks, let's test that on prod.

Issue #57, opened 2021-09-28
Solution from Wiki redirect  


Psi25Omega:
Currently, when we click on the “Get problem solution from Wiki” in the ARCH page, it’ll redirect us to the problem wiki page with editing options.
We can make it such that it redirects just to the wiki page by changing the url wiki-problem in wikihaxx/views.py…
Ideally a quick fix but not sure whether it’s supposed to work the way it’s right now (I.e. the edit problem page)

vEnhance:
The method to change is `wiki_redirect` in `wikihaxx/views.py`

Currently, it only redirects to the "edit" view for pages that are essentially blank (still says "this is an automatically generated article")

In principle we could change it if people are too lazy to edit the page even when newly creating it.

PR #66, opened 2021-10-19
Update views.py

Psi25Omega:
#57 made arch redirect to view problem in wiki instead of edit when a new article is auto-generated


vEnhance:
Thanks. Who are you? (For the spades bounty.)

Psi25Omega:
> Thanks. Who are you? (For the spades bounty.)

Bahnhofstrasse#8974

PR #72, opened 2021-11-07
Update dash/meters/diamonds.html

Psi25Omega:
On the stats page, the form/input bar to enter diamonds is displayed after the list of diamonds you have achieved. Presumably, for people who have found a lot of diamonds, it takes some effort to scroll down. It would be more convenient to move the input bar above the list which has been done in the PR


vEnhance:
Looks good, thanks.

Issue #67, opened 2021-10-21
Feedback on quizzes


vEnhance:
Probably just a text box on the website, corresponding to a 1-1 model against
quiz attempts or something like that? Commiserate about arithmetic errors and
the like.

vEnhance:
Closing due to lack of interest.

PR #73, opened 2021-11-10
Feedback on quizzes #67

Psi25Omega:
This PR addresses issue #67 
It adds a `Feedback` model in `exams/models.py` and links it to a form in `exams/forms.py` which is fetched

**Things that are incomplete/questionable as of now:**
* The post url for the Feedback form in `exams/templates/exams/feedback_form.html`
* The model doesn't have a `OneToOneField` as requested
* The code in `views.py` corresponding to this is crappy
* The addition of `Feedback` model to `admin.py` is uh (?)

**Positives:**
Room for CSS and HTML modifs


vEnhance:
Needs a lot of work, but I left some initial comments for things to fix. I'm afk so I won't be able to test it locally myself for a bit.

Have you got local server working and tested this here to see if it works as anticipated?

Psi25Omega:
I haven’t tested it yet coz I’m not sure whether the code is correct and works with our model…I had some doubts which is why I mentioned the incomplete stuff

vEnhance:
OK, so do you want more advice from me at this point, or just want more
time to work on it?

I would recommend getting things working on a local server. That’s the
second best way to get feedback, make sure the coder runs :).

(The first best way is unit tests but I’m happy to write those for you
later on.)
-- 
*Evan Chen (陳誼廷) *[via evanchen at mit.edu]
https://web.evanchen.cc/ || https://www.facebook.com/evanchenmath/

Issue #71, opened 2021-11-02
Clickable emojis in meters


vEnhance:
People seem to have a hard time finding the "View Stats" button and then
clicking through to the details of each of their stats. So maybe we should make
each of the emojis clickable, having it show the detailed page and automatically
expand the relevant menu

Psi25Omega:
A question…As an example, if a person is on the portal’s home page and clicks on the spades icon, should he be redirected to the `stats’ page with the details of the spades thingy or is this supposed to happen in the home page per se

vEnhance:
> As an example, if a person is on the portal’s home page and clicks on the spades icon, should he be redirected to the stats page with the details of the spades thingy

Yes, that's the hope.

AmoleR:
Probably can be done by giving each a unique `id` (say one of them is `some-category-instead`) and redirecting to `https://www.otis.evanchen.cc/this-is-not-the-url-but-i-forgot-it#some-category-instead`.

vEnhance:
Anone want to take a stab at this? I'm currently being held somewhere against my will so I don't ave my usual work rig to to do as much as I like to do.

TheCoolDinosuar:
I think I'll try, but I'm not sure how long it'll take for me to get everything figured out.

PR #79, opened 2021-12-04
Clickable emojis in meters #71

TheCoolDinosuar:
I have no idea what I'm doing but it is somehow functional.
First, I made the emojis in the meters/small.html be links
Then I added some js at the bottom of stats.html to automatically expand the accordion but it looks kinda scuffed


vEnhance:
Thanks! Who are you? (For spades bounty.)

TheCoolDinosuar:
Yay! I'm Alan Duan

Psi25Omega:
wow that is a hack

PR #82, opened 2021-12-29
hiii

SurivoNan:
i need this charisma :D


SurivoNan:
where's the diamond :C

AmoleR:
What is this even trying to fix? I may be able to review if you specify.

vEnhance:
I think just looking for diamonds and not actually trying to fix anything.

Issue #51, opened 2021-09-08
Incorporate unit artwork into catalog


vEnhance:
Probably just an optional image field to add

Psi25Omega:
i wanna try implementing this
would you like the image to correspond to the `UnitGroup` model or the `Unit` model?

vEnhance:
I was thinking `UnitGroup`

vEnhance:
Closing for now since it doesn't look like I'll have much time for art anyways.

PR #86, opened 2022-01-22
#51 artwork into unit catalog

Psi25Omega:
#51 tl;dr: I coded some lines in and I'm not sure why it's not working.

Um yeah so here's what I did...
1. Added 'artwork', an ImageField to the 'UnitGroup' model
2. The extra code in `core/synopsis.html`...Totally not inspired by the `diamond_list.html`

I'm attaching screenshots from the admin panel and the catalog.
image
image


vEnhance:
So like... can you make it pass the tests?

Psi25Omega:
Ummm yea…?
The problem is I can’t commit from my Linux computer and hence need to edit files one by one
The tests failed on one of many stupid such edits

PR #95, opened 2022-08-23
Diamond leaderboard

TheCoolDinosuar:
I guess it's functional? Currently it redirects staff to the admin leaderboard, because I didn't know what to do if the user wasn't a student because the public field is under Student. I briefly tried UserProfile, but I only wanted students on the lb and I didn't understand the relationship between User and Student, so I just stuck with it. If you have any ideas on how to make it better, then I'd gladly try to implement them


vEnhance:
Pushed a few cosmetic commits, but some more comments on individual lines incoming...

dashboard/views.py
```diff
@@ -154,6 +154,45 @@ def get_queryset(self) -> QuerySet[Achievement]:
 		).order_by('-obtained', '-num_found')
 
 
+class AchievementLeaderboard(
+	LoginRequiredMixin, SuccessMessageMixin, UpdateView[Student, BaseModelForm[Student]]
```
vEnhance, line 158: I don't see why `SuccessMessageMixin` is needed here

dashboard/views.py
```diff
@@ -154,6 +154,45 @@ def get_queryset(self) -> QuerySet[Achievement]:
...
+		students = Student.objects.filter(semester__active=True)
+		rows = get_student_rows(students)
+		rows.sort(key=lambda row: (
+			-row['diamonds'],
+			row['student'].name.upper(),
+		))
+		return rows[0:25]
+
+	def get_object(self, queryset: QuerySet[Student] = None) -> Student:
+		student = Student.objects.get(user=self.request.user)
```
vEnhance, line 186: This line is probably the biggest issue. `Student.objects.get(...)` will expect there to be exactly one return object. So someone who is logged in with no student account (e.g. me) will get a 500 here. Similarly, returners will get a 500 here because they have > 1 student object attached.

You can fix the latter issue by also requiring `semester__active=True` but I'd have to think about what to do with the former: in that case there should be no form, so using `UpdateView[Student, BaseModelForm[Student]]` is sort of painting ourselves into a corner.

Honestly it might be easier to ditch the class-based view and do things by hand here if the form is going to be optional? idk.

dashboard/views.py
```diff
@@ -154,6 +154,45 @@ def get_queryset(self) -> QuerySet[Achievement]:
...
+		return student
+
+	def dispatch(self, request: AuthHttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase:
```
vEnhance, line 189: I think this should be `HttpRequest` instead of `AuthHttpRequest` to match the signature of the base method.

dashboard/views.py
```diff
@@ -154,6 +154,45 @@ def get_queryset(self) -> QuerySet[Achievement]:
...
+		messages.success(self.request, f"Updated settings for {self.object.user.username}!")
+		form.instance.user = self.request.user
+		return super().form_valid(form)
+
+	def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
+		context = super().get_context_data(**kwargs)
+		context['rows'] = self.get_queryset()
+		return context
+
+	def get_queryset(self) -> List[Dict[str, Any]]:
```
vEnhance, line 176: You should use a name other than `get_queryset`. This is an `UpdateView`, there is no queryset attached to it, so this will interact with `SingleObjectMixin` in an undesirable way.

Alternatively, you don't really even need this method, since you only use it once in `get_context_data`, you can just fold it in to the above.

dashboard/views.py
```diff
@@ -154,6 +154,45 @@ def get_queryset(self) -> QuerySet[Achievement]:
...
+):
+	model = Student
+	fields = ('public', )
+	success_url = reverse_lazy('achievement-leaderboard')
+	object: Student
+	template_name = 'dashboard/achievement_leaderboard.html'
+
+	def form_valid(self, form: BaseModelForm[Student]):
+		messages.success(self.request, f"Updated settings for {self.object.user.username}!")
+		form.instance.user = self.request.user
```
vEnhance, line 168: Add `assert isinstance(self.request.user, User)` above this to pass the type check.

roster/models.py
```diff
@@ -112,6 +112,11 @@ class Student(models.Model):
 	enabled = models.BooleanField(
 		default=True, help_text="Allow student to submit/request units."
 	)
+	public = models.BooleanField(
```
vEnhance, line 115: I would have named this something longer since it's not clear to an outsider what `public` should mean on a student object (maybe, `show_on_leaderboard`?). Might be a pain to change now though, so I can also just do it myself later.

vEnhance (changes requested):
Overall, looks cool. Here's some specific issues. The thorniest one is what to do if there is not exactly one matching Student.

vEnhance:
Closing for book-keeping. Feel free to request a re-open if you want to continue working on this.

PR #102, opened 2022-09-28
add float level meters

jeffarjeffar:
Personally I like decimals better, and the text is a bit too close to the emoji anyways
image

Maybe we could add a display setting for this.


vEnhance:
Clubs is always an integer, right? I don't think that one needs to be a float (like diamonds).

vEnhance (approved):
Cool, let's give this a spin. (Also lol .vscode.)

PR #131, opened 2022-11-28
Change the AoPS link

lvj-dev:
Changed the button which leads to the AoPS thread on the problem. In the past it included the entire link (it was shown) and this was ugly and long (leading the button to sometimes even jump off the page). Anyways the user has no use of seeing the link so why no change it to say "Visit AoPS"

small enhancement

Lasitha Vishwajith Jayasinghe


vEnhance:
(1) The links are mostly AoPS but not exclusively so. That means it's more informative to have a link, even truncated, than just "Visit AoPS" button, I believe.

(2) Unrelatedly, you should merge in the most recent changes from the `main` branch into `patch-4` before submitting a pull request. You're missing a few commits.

Issue #117, opened 2022-11-18
Trim long prompts in market admin


vEnhance:
Another easy warm-up task for those who want to ease in:

![image](https://user-images.githubusercontent.com/3750940/202806471-2ad3bf3d-5a98-465f-9ff5-851657912728.png)

Right now, in https://otis.evanchen.cc/admin/markets/market/ the entire prompt of all the markets is displayed. This prompt is sometimes so long that it makes the table readable.

1. Write a `@property` on the market model called `prompt_truncated` that prints the prompt verbatim if it's less than 200 characters, and otherwise cuts off after the 200th character and then appends ellipsis ("...") at the end.
2. Use this in place of `prompt` in the admin interface.

AryKantKode:
hi can i do this

PR #133, opened 2022-11-29
Truncate market prompts issue #117

TheCoolDinosuar:
For #117, I made the market prompts in admin panel truncated if they were too long. I didn't use `@property`, because I didn't know how to change the name of the heading in the admin view if I did it that way.

otis id is 787


vEnhance:
Awesome, thanks :)

PR #148, opened 2022-12-28
Update pset_queue_list.html

RadialFunc:
Checks if the unit is a hidden unit and changes the information appropriately. Goes along with the update in views.py.

OTIS ID: 968

### DIFF
--- a/dashboard/templates/dashboard/pset_queue_list.html
+++ b/dashboard/templates/dashboard/pset_queue_list.html
@@ -23,9 +23,9 @@
               <a href="{{ pset.get_absolute_url }}">{{ pset.unit.group.name }}<sup>{{ pset.unit.code }}</sup></a>
             {% else %}
               {% if pset.unit.group.hidden %}
-                {{ pset.unit.group.name }}<sup>{{ pset.unit.code }}</sup>
-              {% else %}
                 Redacted<sup>❓</sup>
+              {% else %}
+                {{ pset.unit.group.name }}<sup>{{ pset.unit.code }}</sup>
               {% endif %}
             {% endif %}
           </td>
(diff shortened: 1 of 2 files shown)


dashboard/templates/dashboard/pset_queue_list.html
```diff
@@ -22,7 +22,11 @@
             {% if pset.student.user == request.user or request.user.is_staff %}
               <a href="{{ pset.get_absolute_url }}">{{ pset.unit.group.name }}<sup>{{ pset.unit.code }}</sup></a>
             {% else %}
-              {{ pset.unit.group.name }}<sup>{{ pset.unit.code }}</sup>
+              {% if pset.unit.code == 'BKV' or pset.unit.code == 'DKV' or pset.unit.code == 'ZKV' %}
```
Incompleteusern, line 25: Surely there's a better check possible?

RadialFunc: This seemed the easiest to me as I was not sure of what other attributes the pset type object had.

vEnhance: @RadialFunc `core/models.py` is what you want if you want to see what other attributes there are.

In this case you want `pset.unit.group.hidden`

vEnhance:
Cool, now just delete the line in #149 and it should work

Incompleteusern:
image
Logic seems inverted.

Also I wonder if "redacted" is the most appropriate terminology, something more fantasy might be better

RadialFunc:
Pretty sure the value is set to False if the unit is hidden, as originally, views.py filtered out those units.

vEnhance:
> Pretty sure the value is set to False if the unit is hidden, as originally, views.py filtered out those units.

No, I think @Incompleteusern is correct and the logic is inverted.

The Django `filter` method selects things for which all the things are *true*. It means "filter for", not "filter out" (the latter is `exclude`).

(Also, it would be absolutely idiotic for a variable named `hidden` to be false for hidden units. If that ever happens, please slap whoever chose the variable name. :stuck_out_tongue_winking_eye: )

RadialFunc:
Should be good now.

vEnhance:
Thanks, have 5 spades.

PR #149, opened 2022-12-28
Update views.py

RadialFunc:
Doesn't hide hidden units anymore. Goes along with the update in the pset queue html file.

OTIS ID: 968


vEnhance:
> I submitted the pull request, but it seems to be bugging out. 

It's not a bug, it means the one line you commented out failed the style formatter.

Try running `black dashboard/views.py` to fix the issue, it should auto-format.

> Doesn't hide hidden units anymore. Goes along with the update in the pset queue html file.

You should probably submit these in the same pull request, I don't think it makes sense to split them.

Issue #141, opened 2022-12-04
Allow unit petitions to lock a unit but not drop it


vEnhance:
Equivalent to "DROP" followed by "ADD FOR LATER" , but I guess this can be considered a quality-of-life feature or something.

Note that these should be gated by the same rules that "DROP" petitions are.

lvj-dev:
I think the whole point is to stop these requiring confirmation. Like why on earth do we need confirmation for dropping a unit or locking it?

vEnhance:
> I think the whole point is to stop these requiring confirmation. Like why on earth do we need confirmation for dropping a unit or locking it?

People abused them in past by requesting 9 units, downloading them and immediately dropping all 9 to request 9 more

lvj-dev:
But ig no need of any confirmation for locking a unit or do you need?

vEnhance:
still needs confirmation to lock for the same reason that drop needed confirmation

lvj-dev:
So when u lock a unit u get a unit unlocked in place of it?
If so, why don't we make it so that the locking and unlocking confirmation comes in one (to make stuff faster)?

RadialFunc:
working on this.

Incompleteusern:
Should be closed by #155

PR #152, opened 2023-01-02
Update models.py

RadialFunc:
Adds the option to Lock a unit (removing it from unlocked, but stays in your curriculum). Feature #141 .

OTIS ID: 968


Incompleteusern:
nice

vEnhance:
@RadialFunc two notes:

(1) You need to generate a migration for model changes by running `python manage.py makemigrations` (which means for this pull request you won't be able to do it just from the web interface, sorry :P )
(2) Most of the petition processing is done in `aincrad/views.py` so you need to update the code there

vEnhance:
EDIT to add: sorry, `aincrad/views.py` is fine, I forgot it already calls `run_accept`

vEnhance:
wtf how did i lose the commit

vEnhance:
seems I can't push to the remote branch now so I set up #155 instead

Issue #146, opened 2022-12-20
Better one-sem invoice logic


vEnhance:
The current system is super janky and doesn't quite do the right thing if
someone joins for only the spring semester.

vEnhance:
9 spades should be awarded now!

vEnhance:
to elaborate:

i think basically rn if you join for the 2nd semester then the system gets super confused
and thinks you haven't paid for the 1st semester
or smth silly like that
and it would be nice to make the logic correct

Incompleteusern:
I think the issue is that
```python
            else:
                hours_taught = 0
            invoice = Invoice(
                student=student, preps_taught=n, hours_taught=hours_taught
            )
            first_payment_deadline = student.semester.first_payment_deadline
            if first_payment_deadline is not None and first_payment_deadline <= (
                grace_deadline := timezone.now() + timedelta(days=7)
            ):
                invoice.forgive_date = grace_deadline
```

`first_payment_deadline` is set but `most_payment_deadline` is not, so the `most_payment_deadline` gets flagged

Incompleteusern:
Actually the issue is that `grace_deadline` is completely ignored by the rendering

PR #154, opened 2023-01-15
Add Semester Invoice Amount Switch Time

Incompleteusern:
Handles #146.

This supersedes the `num_preps` in `RegistrationContainer` with `one_semester_date` in `Semester` that is checked against the current date right now. Note that the ability to set `num_preps` to more exotic values (\ge 3) is lost.

USER ID: 903, username should be same


vEnhance:
> Handles #146.

Tip: if you write "closes #146" or similar, then github will automatically link the issue to the pull request

vEnhance:
This looks great, thanks so much.

Issue #137, opened 2022-12-02
Bring suggestions/views.py coverage up to > 95%


vEnhance:
Right now there are barely any tests at all I think.

PR #156, opened 2023-01-17
Create Suggestion Queue

### LINKED ISSUE #153: Suggestion queue
Apparently people like watching the backlog move, so if someone wants to
implement a queue for suggestions like with the problem sets, that might be a
welcome addition.

And then maybe I'll have more motivation to not let the suggestions just sit
there for months on end, LOL. (Especially now there are LaTeX requirements,
which should make the process much much smoother on my end.)

Incompleteusern:
Fixes #153 

This adds a basic suggestion queue based off the pset queue.

image

Unacknowledged suggestions are only visible to staff and the submitter.
image

### DIFF
--- a/suggestions/templates/suggestions/problemsuggestion_form.html
+++ b/suggestions/templates/suggestions/problemsuggestion_form.html
@@ -4,10 +4,10 @@
 {% endblock title %}
 {% block main-class %}
   col-12
-{% endblock main-class %}
+{% endblock main-class %}List all
 {% block layout-content %}
   <p>
-    <a href="{% url "suggest-list" %}">⬅️ List all suggestions by you.</a>
+    <a href="{% url "suggest-list" %}">⬅️ List all suggestions by you.</a> Also check out the <a href="{% url "suggestion-queue-listing" %}">submission queue</a>.
   </p>
   <p>
     Thank you for suggesting problems for OTIS!
--- a/suggestions/templates/suggestions/problemsuggestion_form.html
+++ b/suggestions/templates/suggestions/problemsuggestion_form.html
@@ -4,10 +4,10 @@
 {% endblock title %}
 {% block main-class %}
   col-12
-{% endblock main-class %}List all
+{% endblock main-class %}
 {% block layout-content %}
   <p>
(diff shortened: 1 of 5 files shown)


suggestions/templates/suggestions/problemsuggestion_form.html
```diff
@@ -4,10 +4,10 @@
 {% endblock title %}
 {% block main-class %}
   col-12
-{% endblock main-class %}
+{% endblock main-class %}List all
```
vEnhance, line 7: Is this a mistake?

suggestions/templates/suggestions/suggestion_queue_list.html
```diff
@@ -0,0 +1,56 @@
...
+        <td class="text-end">#</td>
+        <td>Suggestion</td>
+        <td>Weight</td>
+        <td>Unit</td>
+        <td>Timestamp</td>
+        <td class="text-end">PK</td>
+      </tr>
+    </thead>
+    <tbody>
+      {% for suggestion in object_list %}
```
vEnhance, line 20: I think it'd be nice to have `context_object_name = "suggestions"` in `views.py` so we can use `suggestions` here instead of `object_list`, which I think reads more nicely

suggestions/urls.py
```diff
@@ -11,4 +11,7 @@
     ),
     path(r"<int:pk>/", views.ProblemSuggestionUpdate.as_view(), name="suggest-update"),
     path(r"list/", views.ProblemSuggestionList.as_view(), name="suggest-list"),
+    path(
+        r"queue/", views.SuggestionQueueList.as_view(), name="suggestion-queue-listing"
```
vEnhance, line 15: Seems it'd be better to name it `suggest-queue-listing` to be consistent with the other URL's

vEnhance (changes requested):
Looks nice, here are a few minor comments.

vEnhance:
Also a few simple sanity checks in `tests.py` would be appreciated.

Incompleteusern:
> Also a few simple sanity checks in `tests.py` would be appreciated.

I think this falls under https://github.com/vEnhance/otis-web/issues/137 and would rather not do tests right now

Incompleteusern:
Also sorry for bothering you in multiple places about this, but please check my discord dms with you.

vEnhance (approved):

vEnhance:
Thank you! It'll be live soon.

PR #157, opened 2023-01-17
Grace Invoice Changes

Incompleteusern:
Actually deals with #146 I think.

Render warnings and info for grace deadline as appropriate. This also merges `is_delinquint` and `is_payment_locked` using `-3`, `-2`, `-1` respectively as returned statuses.

image
image

Other changes is setting the grace deadline as the maximum of `most_payment_deadline` and seven days from now, in the rare case that someone is approved after `first_payment_deadline` but before `most_payment_deadline`

Some changes to the words themself have occurred, override as necessary.

ID: 903

### DIFF
--- a/roster/admin.py
+++ b/roster/admin.py
@@ -348,7 +348,9 @@ def build_students(queryset: QuerySet[StudentRegistration]) -> int:
             if first_payment_deadline is not None and first_payment_deadline <= (
                 grace_deadline := timezone.now() + timedelta(days=7)
             ):
-                invoice.forgive_date = max(grace_deadline, student.semester.most_payment_deadline)
+                invoice.forgive_date = max(
+                    grace_deadline, student.semester.most_payment_deadline
+                )
             invoices_to_create.append(invoice)
         Invoice.objects.bulk_create(invoices_to_create)
     return count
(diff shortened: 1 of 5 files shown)


roster/admin.py
```diff
@@ -348,7 +348,9 @@ def build_students(queryset: QuerySet[StudentRegistration]) -> int:
             if first_payment_deadline is not None and first_payment_deadline <= (
                 grace_deadline := timezone.now() + timedelta(days=7)
             ):
-                invoice.forgive_date = grace_deadline
```
Incompleteusern, line 351: revert this change if seen as undesirable

vEnhance:
So I thought about this and I think do not want the grace deadline to be visible to the end-user (this is what I had decided earlier too), especially since it adds a lot of extra complexity to the code.

The reason is that the grace deadline is meant to be for my internal use; if someone has some good reason they are late on the deadline (e.g. just joined), then I want to give them a bit of extra time. But they are still *late*, so from a user's perspective that's what I want the website to show, I just don't want to hard-lock them out.

To clarify, my goal for #146 was that I wanted to refactor the code somehow so that people joining in the narrow window between the end of 1st and 2nd semester shouldn't _need_ to have a grace deadline. Basically, if you join, say, December 25, then you shouldn't need to pay anything until January 21. So what I had in mind was to flag people who joined for only the spring so that `first_payment_deadline` did not apply to them at all. This shouldn't IMO be the grace deadline's job.

Sorry for not being clear what I was going for.

Incompleteusern:
Where is the January 21 specified in code? Is `first_payment_deadline` and `most_payment_deadline` manually switched by you?

PR #161, opened 2023-01-24
Add created_at for late students

Incompleteusern:
Yet another attempt at #146 

This makes it so that every invoice has a `created_at` (migrations use Jan 1 1970) used to see which deadlines apply to the invoice. Note that this requires the payment deadline to be updated to take effect. 

Assuming that `first_payment_deadline` and `most_payment_deadline` haven't been updated to this year yet, this might pose a migration issue for users who have joined in 2023. This also merges `is_delinquent` and `payment_locked` into one


vEnhance:
Can we leave `forgive_date` as a functional field in the model? I think this is an improvement, but there are still going to be edge cases that I need to manually deal with (e.g. "I can pay you in a month when so-and-so family situation resolves"), so broadly speaking having no manual overrides is not really an option.

As for the migration, one thing we could do for invoices with no `created_at` is to copy the `created_at` field from `StudentRegistration` and use that as a good-enough-for-now approximation. That's a complicated manual migration though so I can do that myself.

roster/tests.py
```diff
@@ -108,35 +108,25 @@ def test_delinquency(self) -> None:
...
             self.assertEqual(alice.payment_status, 1)
-            self.assertFalse(alice.is_payment_locked)
             self.assertFalse(alice.is_delinquent)
         with freeze_time("2022-09-25", tz_offset=0):
             self.assertEqual(alice.payment_status, 2)
-            self.assertFalse(alice.is_payment_locked)
             self.assertFalse(alice.is_delinquent)
         with freeze_time("2022-10-15", tz_offset=0):
             self.assertEqual(alice.payment_status, 3)
-            self.assertTrue(alice.is_payment_locked)
```
vEnhance, line 132: can we add these lines back?

roster/views.py
```diff
@@ -248,7 +248,6 @@ class UpdateInvoice(
         "adjustment",
         "extras",
         "total_paid",
-        "forgive_date",
```
vEnhance, line 251: same with these

roster/tests.py
```diff
@@ -108,35 +108,25 @@ def test_delinquency(self) -> None:
...
+            invoice: Invoice = InvoiceFactory.create(
+                student=alice,
+                preps_taught=2,
+            )
 
         # Alice has paid $0 so far
         self.assertEqual(invoice.total_owed, 480)
         with freeze_time("2022-09-05", tz_offset=0):
             self.assertEqual(alice.payment_status, 4)
-            self.assertFalse(alice.is_payment_locked)
```
vEnhance, line 120: is there a reason the `alice.is_payment_locked` tests were deleted?

vEnhance (changes requested):

roster/tests.py
Incompleteusern: I merged `is_payment_locked` and `is_delinquent` in this pr. I can undo if desired

Incompleteusern:
Rebasing mangled the pr

roster/tests.py
vEnhance: I guess that's fine

vEnhance (approved):

vEnhance:
```python
def set_invoice_created_at_by_student_reg(apps: Any, scheme_editor: Any):
    Invoice = apps.get_model("roster", "Invoice")
    Invoice.objects.filter(student__reg__isnull=False).update(
        created_at=Subquery(
            Invoice.objects.filter(pk=OuterRef("pk")).values(
                "student__reg__created_at"
            )[:1]
        )
    )
```

Man, subqueries still make my head spin. You'd think we could just do `.update(created_at=F("student__reg__created_at"))` or something...

PR #163, opened 2023-01-25
Add diamond comparison

Incompleteusern:
This allows people to compare their diamonds with other people's diamonds for sharing purposes ~~trading diamonds~~

For users A, B and the following diamonds:
- "Both Have" is owned by both
- "I Have" is owned by A
- "They Have" is owned by B
- "Neither Have" is owned by Neither

The new colors are light blue which displays diamonds that the viewer has but the viewee does not and orange for those that the viewee has but the viewer does not. Creator diamond info is not displayed while comparing.

No share, A's view: 
image
With share from B, A's view:
image

The code is very goofy but this whole idea is goofy to be honest

### DIFF
--- a/env
+++ b/env
@@ -31,7 +31,7 @@ IS_PRODUCTION=0
 # ADDRESS_REGULAR="my boston address"
 # ADDRESS_HOME="my home address"
 
-# CERT_HASH_KEY="cert hash key"
+CERT_HASH_KEY="cert hash key"
 
 # STRIPE_PUBLISHABLE_KEY="pk_test_..."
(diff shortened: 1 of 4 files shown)


vEnhance:
lmao, why not

env
```diff
@@ -31,7 +31,7 @@ IS_PRODUCTION=0
 # ADDRESS_REGULAR="my boston address"
 # ADDRESS_HOME="my home address"
 
-# CERT_HASH_KEY="cert hash key"
```
Incompleteusern, line 34: fuck

vEnhance: lol

PR #165, opened 2023-01-25
Fix bugs with diamond comparison pr

Incompleteusern:


vEnhance:
I actually fixed this myself already in 3fb9af82.

Basically I think everything should be done by user pk rather than student pk.

Incompleteusern:
Yeah that's better

PR #168, opened 2023-01-27
Increase Verbosity for already unlocked diamond messages

Incompleteusern:
image

There probably is a better replacement than a sob emote, feel free to suggest alternatives :).


Incompleteusern:
Reverted from lock back to sob emoji

PR #171, opened 2023-02-01
Suggestion Deletion

### LINKED ISSUE #170: Add way to delete suggestions

Incompleteusern:
Fixes #170 

This adds a way to delete created suggestions with a button 
image

However, do note that certain parts of this pr are very scuffed, and complain for a better implementation if glaring enough
- Redirecting with the button is handled by js
- The button is placed in the same table entry with the disable tex button, with a hackily added `margin-left` to add some distance between the two of them

### DIFF
--- a/dashboard/tests.py
+++ b/dashboard/tests.py
@@ -430,7 +430,7 @@ class TestPSet(EvanTestCase):
 
         upload = UploadedFile.objects.filter(benefactor=alice, unit=unit).first()
 
-        assert upload != None
+        assert upload is not None
 
         self.assertTrue(upload.unit == unit)
         self.assertTrue(upload.owner == alice.user)
@@ -449,7 +449,7 @@ class TestPSet(EvanTestCase):
 
         upload = UploadedFile.objects.filter(pk=pk).first()
 
-        assert upload != None
+        assert upload is not None
 
         self.assertTrue(upload.description == "bark")
 
--- a/dashboard/tests.py
+++ b/dashboard/tests.py
@@ -430,6 +430,8 @@ class TestPSet(EvanTestCase):
 
         upload = UploadedFile.objects.filter(benefactor=alice, unit=unit).first()
 
+        assert upload != None
+
         self.assertTrue(upload.unit == unit)
         self.assertTrue(upload.owner == alice.user)
 
@@ -446,6 +448,8 @@ class TestPSet(EvanTestCase):
         )
 
         upload = UploadedFile.objects.filter(pk=pk).first()
+
+        assert upload != None
 
         self.assertTrue(upload.description == "bark")
 
--- a/suggestions/templates/suggestions/problemsuggestion_form.html
+++ b/suggestions/templates/suggestions/problemsuggestion_form.html
@@ -120,8 +120,8 @@
         <td>
           <button id="render-toggle" class="btn btn-info">Disable instant TeX</button>
           {% if pk %}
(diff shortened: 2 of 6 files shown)


suggestions/templates/suggestions/problemsuggestion_form.html
```diff
@@ -119,7 +119,10 @@ <h1 class="alert-heading">Line length exceeded in solution!</h1>
         </td>
         <td>
           <button id="render-toggle" class="btn btn-info">Disable instant TeX</button>
-        </td>
+          {% if pk %}
+            <button id="delete" class="btn btn-warning" style="margin-left: 50px">Delete Suggestion
+            </button>
+          {% endif %}
```
vEnhance, line 125: you dropped a `</td>`

vEnhance: Why not just put the button in a different table `td`?

suggestions/templates/suggestions/problemsuggestion_form.html
```diff
@@ -224,6 +227,13 @@ <h3 class="alert-heading">
             render_solution();
         });
 
+        {% if pk %}
+          $("#delete").click(function(e) {
+            e.preventDefault();
+            window.location = '{% url 'suggest-delete' pk %}';
+          })
+        {% endif %}
+
```
vEnhance, line 236: This is pretty bad, I think you should just use a link rather than a `<button>`.

You can make it look like a button using `btn-link` class, but the underlying element should be `<a>` rather than this hack.

suggestions/templates/suggestions/problemsuggestion_form.html
```diff
@@ -234,6 +244,7 @@ <h3 class="alert-heading">
             }
             return true;
         }
+
```
vEnhance, line 247: revert this?

suggestions/views.py
```diff
@@ -92,7 +93,22 @@ def get_context_data(self, **kwargs: Any):
             raise PermissionDenied("Please log in.")
         if not (self.request.user.is_staff or self.request.user == self.object.user):
             raise PermissionDenied("Logged-in user cannot view this suggestion")
+        
+        context["pk"] = self.kwargs.get("pk", None)
         return context
+        return context
```
vEnhance, line 99: duplicated `return context`, the linter should probably catch unreachable statements like this, tbh

vEnhance (changes requested):

vEnhance:
BTW, `black` got updated to `v23` a few hours ago, so I'll need to update the `main` branch to reflect this

so you'll want to merge that in to pass the automated github ci

suggestions/templates/suggestions/problemsuggestion_form.html
Incompleteusern: That runs into an issue with the fact that `colspan=2` is set a bunch of times higher up, which would also need to be modified conditionally I think

dashboard/tests.py
```diff
@@ -447,6 +449,8 @@ def test_file_operations(self):
 
         upload = UploadedFile.objects.filter(pk=pk).first()
 
+        assert upload != None
```
Incompleteusern, line 452: These two new lines is to get pyright to stop complaining

Incompleteusern:
I've expanded the suggestion delete message to be based off the hint delete message

image

vEnhance (approved):

PR #174, opened 2023-02-08
Add Niceness Rating

Incompleteusern:
Feel free to reject/request changes if this pr doesn't really suit OTIS in its philosphy

image
image


Incompleteusern:
This has been rebranded as niceness

arch/models.py
```diff
@@ -46,6 +49,41 @@ def get_absolute_url(self):
...
+
+    problem = models.ForeignKey(
+        Problem,
+        on_delete=models.CASCADE,
+        help_text="The container of the current vote.",
+    )
+
+    niceness = models.FloatField(
+        help_text="A student submitted number from 0 to 10 used to indicate the approximate niceness of a problem.",
+        validators=[MaxValueValidator(50)],
```
vEnhance, line 84: should be out of 10 right?

arch/models.py
```diff
@@ -46,6 +49,41 @@ def get_absolute_url(self):
     def get_statement(self) -> Optional[str]:
         return get_disk_statement_from_puid(self.puid)
 
+    @property
+    def niceness(self) -> float:
+        total: int = 0
+
+        votes = self.vote_set.all()
```
vEnhance, line 56: might want to add a type hint for `vote_set` into the `Problem` class so this doesn't fail the type check

arch/models.py
```diff
@@ -46,6 +49,41 @@ def get_absolute_url(self):
...
+
+        if len(votes) > 0:
+            for vote in votes:
+                total += vote.niceness
+
+            total /= len(votes)
```
vEnhance, line 62: i would rather just `return round(total / len(votes), 2)` here, shorter and doesn't cause a type clash

arch/models.py
vEnhance: shouldn't this be a small positive integer field actually? rather than a floatfield

right now -9000.1434 is a valid input

arch/models.py
vEnhance: aren't you dividing by len(votes) twice now?

seems like you should move the subsequent `        return round(total / len(votes), 2)` here and just delete this line

vEnhance (changes requested):
Looks mostly fine, just `len(votes)` divided twice I think

vEnhance (approved):

vEnhance:
I merged in the recent `main` push and did some cleanup:

- I still think the division code was needlessly complicated, so I rewrote it in 7361f22.
- The `VoteForm` seemed completely useless to me? As far as I can tell it is never used, and the type hint should be `BaseModelForm[Vote]`. See b2a4b87.

Can you sanity-check these for me and test that things still work as intended on your local dev?

vEnhance:
I'm also going to add `created_at` and `updated_at` so that I have some semblance of time tracking (I think looking back I really should have these by default on every model that's editable by users).

Incompleteusern:
still seems to work as intended

vEnhance:
Cool, thanks.

PR #177, opened 2023-02-12
Fix

Incompleteusern:
This fixes aincrad's api from throwing a 500, and replaces the☕emoji which is usually coffee with something else, unless people use teapots for coffee?


vEnhance:
Can you run black v23 and push? That should make it pass CI.

PR #182, opened 2023-02-23
[92%] Roster Tests Partial Progress

Incompleteusern:
This does not fix the roster/views.py issue, since it only bumps it up to 92%. However, I've determined that the remaining lines that are missing seem difficult or even impossible to do properly so I don't feel like doing them for the new future.

image


vEnhance:
Thanks for being the tests hero :D

PR #185, opened 2023-03-02
add alttext

### LINKED ISSUE #143: Short poems describing each RPG level
Since from like level 72 onwards the level names get fantasy based and stuff, we could write cool poems describing each level. For example level 79 is Necromancer and you could have a poem like:

Through your eternal sleep he creeps,
On your heart his mark he leaves;
Seeping dark into your veins
His call shall permeate your grave.

You can't hide, and you can't run
Not from the Necromancer, not when he's done.

Of course, we don't have to restrict ourselves only to level 72 onwards, it's possible to write short poems for the earlier levels as well. As for who writes the poems, it could be like everyone interested can submit poems for each level and then the OTIS overlord could see which poem is the best. Maybe there could be this new button called "Lore" between the Submit Unit and Past Work which takes to a page containing the poem. 

Note: Doing this on the wiki would actually be pretty cool; but the wiki is publicly available and people would know the names of levels they have not yet reached; would feel like a major spoiler.

Incompleteusern:
Fixes #143.

This adds a very basic alt-text as suggested by @Novus-Stellachan

image


vEnhance:
Switched to `linebreaks` instead of the ugly `white-space:pre;` and removed the `<hr/>`.

Now I have to decide how to populate the poems...

PR #186, opened 2023-03-04
Add Test Data for Environment

### LINKED ISSUE #78: Make convincing looking test data
For testing, it would if we could, by running a script, generate ~1k students working on ~200 units some of which are unlocked, locked, approved. Also would make writing tests cases way easier.

Incompleteusern:
Fixes #78.

Adds external script `populate.py` to create some testing data, a bit hacky. Feel free to modify as necessary.

image
image


vEnhance:
I think this might be too slow right now since every object created is a separate database query.

I'll try to rewrite using `create_batch` I guess

Incompleteusern:
`bulk_create` might be well suited for that purpose too ig

vEnhance:
i never want to see this code again

PR #190, opened 2023-03-07
fix(js): update js through files

AmoleR:
Updated javascript syntax and made a few modifications through all files with important JS files.

Also bumped JQuery 3.2.1 to JQuery 3.6.3. Can someone check to make sure that nothing is broken with the site?


vEnhance:
Thanks.

PR #193, opened 2023-03-07
Use bulk operations

### LINKED ISSUE #192: Use `bulk_create` in `populate.py`
As said, `create_batch` doesn't actually speed things up, use `bulk_create` when it can be used, which is almost all cases

Incompleteusern:
Fixes #192 

This dramatically speeds up data generation time through bulk operations
image
image


Incompleteusern:
fixtures application took 13 seconds

vEnhance:
Sweet, thanks much.

PR #198, opened 2023-03-12
Add way to hide Solution Page

### LINKED ISSUE #197: Allow for User Diamonds to Decide whether the Solution/Hex code should be Accessible
Willing to work on this if its decide to be allowed.

Incompleteusern:
Fixes #197.

This allows users (dotted) to hide their diamond solution page.

image
image


vEnhance:
Cool, thanks.

PR #202, opened 2023-03-28
User Markets (Fine To Close)

Incompleteusern:
This is an attempt to implement user markets, please feel free to close if its too much of a hassle to maintain in the long run, as discussed in the discord. 

The major difference here from discussion in the discord is that this treats user made markets the same as Evan made markets, namely in counting for spades and taking a position in the queue. This should be insignificant in events. This also disables answer modification after one week.

This ended up being a decent bit less hacky than I thought it would be, though still hacky.

image
image
image
image


Incompleteusern:
image
the spacing is fixed

vEnhance:
I still think this is too much code complexity for a fairly rarely used case.
Also, I think in practice, I would want to moderate/review the markets anyways and schedule them myself.

Incompleteusern:
Fair enough.

PR #204, opened 2023-04-02
Unit Petition Logic Changes

### LINKED ISSUE #203: Option to drop locked units from curriculum
This shouldn't require Evan's permission, right?

Incompleteusern:
Fixes #203 

Also attempts to streamline inquiry logic through moving logic to a separate function while staying the same afaik, tests still pass,

It might be best to review this commit by commit.

### DIFF
--- a/roster/tests.py
+++ b/roster/tests.py
@@ -317,6 +317,42 @@ class RosterTest(EvanTestCase):
         self.assertEqual(alice.unlocked_units.count(), 6)
 
         self.login(firefly)
+        self.assertHas(
+            self.post(
+                "inquiry",
+                alice.pk,
+                data={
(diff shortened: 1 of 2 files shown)


Incompleteusern:
Also this might require more extensive testing to see that logic isn't inadvertently broken

Incompleteusern:
Why did it merge all the user market code ;-;

vEnhance:
Can you write some tests in `roster/tests.py` to make sure this logic checks out?

Incompleteusern:
`test_inquiry` tests are already decently robust, it detected the two times I messed up the staff check and another time.

Incompleteusern:
Will write tests for this logic though

Issue #206, opened 2023-04-13
Catalog as a Checklist?


Incompleteusern:
https://otis.evanchen.cc/core/catalog/ could color complete units into a different color.

Something like this
![image](https://user-images.githubusercontent.com/58920010/231849911-a42aecbb-3912-4a1f-88a0-d2d9c75ecded.png)

vEnhance:
Seems like a reasonable feature suggestion.

Incompleteusern:
Going to note that this is partially completed with a checkmark, but the granular level of versions of units is not done

jatloe:
A small extension: Using a third color for units that you AK?

vEnhance:
server doesn't have a way of knowing whether you AK'd a unit because the materials stuff is all done on my machine (so e.g. server doesn't know what problems are on each unit or how much they're worth, it just has the pdf handout)

Incompleteusern:
You can AK all units except for like secret units and any other unit with <= 2 problems iirc.

vEnhance:
Right but the point is the server doesn't know _whether_ you AK'ed the unit, 
because it doesn't know the total clubs in the unit (which keeps changing over 
time), hence can't compare it to the number of clubs claimed.

You'd need a Boolean flag on each pset submission or something.

PR #209, opened 2023-04-20
Catalog Checkmarks

Incompleteusern:
Work on #206 but doesn't handle individual Unit checkmarks and only UnitGroup ones.

image


vEnhance:
Seems like you only check against the latest student? I feel like it should check all years, maybe it's just me?

Also, we use `sql_utils` in this project, so the `Exists` subquery can usually be simplified a lot, see https://pypi.org/project/django-sql-utils/. Should be some examples in this repository as well.

Incompleteusern:
I'm going under the assumption that the latest student has all units?

vEnhance:
Thanks!

PR #213, opened 2023-04-28
Prevent Line Check Failing

### LINKED ISSUE #205: Line length Failure Check Appears If Mathjax doesn't Load

Incompleteusern:
Fixes #205 by having the check code run before the render code, so if MathJax is not defined for some reason the check code already has toggled off the failing error before the syntax error occurs. Also rearranges some things and gets rid of the `autorender_enabled && render_enabled` condition.


vEnhance:
Thanks, let's try this out.

Issue #207, opened 2023-04-13
Diamond Wiki Pages


Incompleteusern:
These pages should be createable without embedding the actual diamond code. On that note, wiki diamond might have to change tho idk

PR #214, opened 2023-04-28
Diamond in Wiki Based off pk

Incompleteusern:
A few reasons for this:
- You can't make a wiki page currently with a diamond without leaking it
- The wiki diamond isn't special since there's also a wikipage for Start Here as a result
- More diamond pages would be cool

image
image

The tag is only expanded to show information if the diamond is not user made and at least one person has found the diamond.

### DIFF
--- a/wikihaxx/mdx/otis.py
+++ b/wikihaxx/mdx/otis.py
@@ -42,7 +42,7 @@ class OTISPreprocessor(markdown.preprocessors.Preprocessor):
 
                 if tag_name == "diamond":
                     try:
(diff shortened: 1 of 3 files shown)


Incompleteusern:
Also fixes #207 ig, though if this pr is denied that should also be closed.

Incompleteusern:
It's fine to close this pr if deemed unnecessary/too complex btw.

vEnhance:
We can give this a try.

vEnhance:
I guess this lets you get the name and description of a diamond even if you don't have it unlocked, but the wiki is advertised to contain spoilers anyways.

Issue #219, opened 2023-07-25
Link on Diamond Page


Incompleteusern:
![image](https://github.com/vEnhance/otis-web/assets/58920010/fcced5a7-7655-458c-b2e3-85fb6d5dd7b0)
This could automatically add the url to open the relevant `https://otis.evanchen.cc/rpg/stats/<id>/#headingCharisma`

vEnhance:
seems like a good beginner task :)

PR #227, opened 2023-07-27
Link on diamond page thing

jatloe:
OTIS-WEB username: `jatloe`
Supposed to partially resolve [https://github.com/vEnhance/otis-web/issues/219](https://github.com/vEnhance/otis-web/issues/219)
The issue I have is that I don't know how to infer a student without a try-catch, so there is a try-catch which isn't very nice

### DIFF
--- a/rpg/urls.py
+++ b/rpg/urls.py
@@ -27,7 +27,4 @@ urlpatterns = [
     path(
         r"forge/<int:student_pk>/", views.DiamondUpdate.as_view(), name="diamond-update"
     ),
-    path(
-        r"stats/<int:student_pk>/", views.stats, name="heading-charisma"
-    ),
 ]
--- a/rpg/urls.py
+++ b/rpg/urls.py
@@ -27,4 +27,7 @@ urlpatterns = [
     path(
         r"forge/<int:student_pk>/", views.DiamondUpdate.as_view(), name="diamond-update"
     ),
+    path(
+        r"stats/<int:student_pk>/", views.stats, name="heading-charisma"
+    ),
 ]
--- a/rpg/views.py
+++ b/rpg/views.py
@@ -18,7 +18,7 @@ from django.core.exceptions import PermissionDenied
 from django.db.models import OuterRef, Q
 from django.db.models.query import QuerySet
 from django.forms.models import BaseModelForm
-from django.http import HttpResponse, Http404
+from django.http import Http404, HttpResponse
 from django.http.response import HttpResponseBase
 from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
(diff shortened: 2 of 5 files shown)


jatloe:
WHAT !

jatloe:
what do you MEAN it failed

rpg/urls.py
```diff
@@ -27,4 +27,7 @@
     path(
         r"forge/<int:student_pk>/", views.DiamondUpdate.as_view(), name="diamond-update"
     ),
+    path(
+        r"stats/<int:student_pk>/", views.stats, name="heading-charisma"
+    ),
```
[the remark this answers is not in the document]
vEnhance: That's what it was originally.

I think this approach is fine too, but in that case adding a new URL path named "charisma-heading" is completely unnecessary because it is identical to the existing "stats" view. Just delete all the changes here and use "stats" in the template.

rpg/views.py
```diff
@@ -140,6 +143,10 @@ def get_context_data(self, **kwargs: dict[str, Any]):
             self.request.user.pk, self.amount, settings.CERT_HASH_KEY
         )
         context["pk"] = self.request.user.pk
+        try:
+            context["student_pk"] = infer_student(self.request).pk
+        except:
+            context["student_pk"] = -1
```
vEnhance, line 149: i think you should use `None` instead of `-1`, that's more clear what is meant imo

vEnhance: also you should use explicit `except Http404`, in general using a bare `except` is considered bad

vEnhance (changes requested):
Changes to make:

- Don't create a separate `heading-charism` URL, it's redundant, and just use `stats`.
- Use `None` rather than `-1` for the case where no student exists.
- Use `except Http404` rather than bare `except`.

The unit test (which verifies the logic is correct) is failing because the template you modified is also used in a different view (namely the trading diamonds list), where no student_pk is provided. This will fix itself if you use `None` rather than `-1`, which is the value typically used in the "does not exist" case.

You're also failing some style tests. Maybe I'll add a script to auto-fix style issues.

vEnhance:
if you `git pull --recursive` there's an updated `lint.sh` that will fix some of the style issues for you (instead of merely telling you they exist)

rpg/views.py
jatloe: I don't think Http404 is a BaseException, so it can't be caught? I'm not sure if that's actually how it works

jatloe:
Added the new fixes

jatloe:
ok the :3 was to make the style check happy

jatloe:
let's try this again

jatloe:
please?
:candycorn:

jatloe:
YEEEEEEEEEEEEEEEEEEEESSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
clear

vEnhance (approved):

vEnhance:
thanks, looks good

PR #249, opened 2023-08-08
Twemoji support

### LINKED ISSUE #243: The club symbol looks really different aesthetically in terms of size etc
image

Incompleteusern:
Fixes #243.
image
image


jatloe:
get style checked

jatloe:
wow you failed two of them
~~33.3% F~~

vEnhance:
oof

vEnhance:
let me attempt to do stuff lol

vEnhance:
that should hopefully pass tests now, can you verify it does what it's supposed to

core/templatetags/otis_extras.py
```diff
@@ -46,6 +46,14 @@ def getprofile(user: User) -> Optional[UserProfile]:
         return None
 
 
+@register.filter(name="getconfig")
+def getconfig(user: User, config: str) -> bool:
+    try:
+        return getattr(UserProfile.objects.get(user=user), config)
+    except Exception:
+        return False
```
vEnhance, line 54: in general i think `except Exception` is bad because it makes it hard to catch things

i think it should be like, try to get `profile = UserProfile.objects.get(user=user)`, and if this throws an error return False, else proceed as normal

dashboard/templates/dashboard/portal.html
```diff
@@ -136,29 +136,27 @@ <h1>Getting Started — Setup Instructions</h1>
...
           </div>
-        {% else %}
-          <h1>Welcome back</h1>
-          <p>
-            You are logged in as <code>{{ request.user.username }}</code>.
-          </p>
-        {% endif %}
-      </div>
-    {% endwith %}
+        </div>
```
vEnhance, line 152: wtf was i just missing a `</div>`?

otisweb/templates/base.html
```diff
@@ -25,6 +26,18 @@
           href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/themes/smoothness/jquery-ui.css" />
     <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"></script>
     <script src="{% static "k.js" %}"></script>
+    {% if request.user|getconfig:"use_twemoji" %}
+      <script src="https://unpkg.com/twemoji@latest/dist/twemoji.min.js"
+              crossorigin="anonymous"></script>
+      <style>
```
vEnhance, line 32: we should just put this in `otis.css` rather thar inline on every page

vEnhance (changes requested):

core/templatetags/otis_extras.py
Incompleteusern: The issue is that I don't actually know what exceptions it likes to throw :skull:

dashboard/templates/dashboard/portal.html
Incompleteusern: I think this is just the result of removing `{% endwith %}`

vEnhance: oh i see

core/templatetags/otis_extras.py
Incompleteusern: Should be fixed

vEnhance (approved):

PR #253, opened 2023-08-09
Fix User Group Not Being Added

### LINKED ISSUE #252: User groups are missing in fixtures
For some reason the fixtures with dummy data don't put the generated users into `Verified` or `Active Staff` respectively. Need to find out why, regen the dummy data and then update the `all.json` fixtures file accordingly.

Incompleteusern:
Fix #252

### DIFF
--- a/fixtures/populate.py
+++ b/fixtures/populate.py
@@ -423,14 +423,11 @@ def init():
 
     print(f"Creating {args.stu_num} user accounts")
     # users
-    users: list[User] = fast_bulk_create(
-        UserFactory, args.stu_num, groups=(verified_group,)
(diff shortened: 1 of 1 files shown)


vEnhance:
can you check that the runtime is not too terrible without the optimization

i feel like the "correct" thing to do is fast bulk create the users and try to add them all at once to the Verified group, the latter should be an O(1) operation

Incompleteusern:
Its not terrible, still runs very fast by default. I'll try doing that though.

vEnhance:
OK, can you also commit an updated `all.json` once you're all done

PR #255, opened 2023-08-10
Catalog Checklist completely.

### LINKED ISSUE #206: Catalog as a Checklist?

Incompleteusern:
Fixes #206 .
image

Ignore the horribly mangled git history as a result of my laziness.


Incompleteusern:
Ok pyright wont shut up apparently

Incompleteusern:
Oops ;-;

Incompleteusern:
The construct
```python
                      {% if not forloop.last %}
                        {{ unit.code }},
                      {% else %}
                        {{ unit.code }}
                      {% endif %}
```
is rather stupid but djlint kind of forces it I think

vEnhance:
i think we can just put check marks on the individual units rather than the group if we're going to have that information

also i'm happy to just drop the commas if it's that annoying

let me try that now

vEnhance:
oh i see this is actually super annoying because it's a listview of unitgroup's rather than unit's

uhh

ok maybe i should rewrite it

Incompleteusern:
Feel free to do so. It might end up being more roundabout though since the catalog is split into unit groups which are in turn split into units.

Also hopefully this has O(1) database queries, since appararently django filter on query set doesn't hit the database most of the time.

vEnhance:
ok that should be a lot simpler now

vEnhance:
does that look reasonable?

Incompleteusern:
Yeah this is better. Not a fan of the emojis being in the subject group title (Coloring the text or something would be better IMO)

Incompleteusern:
Yeah this is better. Not a fan of the emojis being in the subject group title (Coloring the text or something would be better IMO)

Incompleteusern:
Codewise, everything is a lot better and cleaner lol. I'll make a poll in #billboards to ask what people prefer ig.

vEnhance:
OK, thanks

Issue #251, opened 2023-08-09
Reduce Twemoji to just known misfunctioning emojis


Incompleteusern:
This would just be spades/clubs/hearts/diamonds or something

Incompleteusern:
![image](https://github.com/vEnhance/otis-web/assets/58920010/32c40869-c0a1-459e-bda1-bf4e53cf1615)

Also, replacing the spade on meters with another emoji about dexterity might be better

PR #262, opened 2023-08-10
Update Twemoji and Fix Diamond Emoji

Incompleteusern:
This restricts twemoji to just emojis used in meters.

It also updates the emoji for the charisma meter to also be in the card suit 2660-266f range instead of the seemingly unrelated "Black Diamond" codepoint.


Incompleteusern:
Fixes #251

Incompleteusern:
Here's an example of the new thing.
image

Incompleteusern:
Also old vs new emoji
image

image

PR #270, opened 2023-08-13
refactoring

menezesd:

Make more use of f-strings, simplify logical expressions and control flow

dean

### DIFF
--- a/aincrad/views.py
+++ b/aincrad/views.py
@@ -193,7 +193,15 @@ JOB_VENUEQ_INIT_KEYS = (
 
 
 def venueq_handler(action: str, data: JSONData) -> JsonResponse:
-    if action == "grade_problem_set":
+    if action == "accept_inquiries":
+        for inquiry in UnitInquiry.objects.filter(
+            status="INQ_NEW",
+            student__semester__active=True,
+            student__legit=True,
+        ):
+            inquiry.run_accept()
+        return JsonResponse({"result": "success"}, status=200)
+    elif action == "grade_problem_set":
         # mark problem set as done
         pset = get_object_or_404(PSet, pk=data["pk"])
         original_status = pset.status
@@ -216,53 +224,11 @@ def venueq_handler(action: str, data: JSONData) -> JsonResponse:
             # remove the old unit since it's done now
             pset.student.unlocked_units.remove(pset.unit)
         return JsonResponse({"result": "success"}, status=200)
-    elif action == "accept_inquiries":
-        for inquiry in UnitInquiry.objects.filter(
-            status="INQ_NEW",
-            student__semester__active=True,
-            student__legit=True,
-        ):
-            inquiry.run_accept()
-        return JsonResponse({"result": "success"}, status=200)
-    elif action == "mark_suggestion":
-        suggestion = get_object_or_404(ProblemSuggestion, pk=data["pk"])
-        suggestion.status = data["status"]
-        suggestion.eligible = data["eligible"]
-        if "staff_comments" in data:
-            if suggestion.staff_comments:
-                suggestion.staff_comments += "\n\n" + "-" * 40 + "\n\n"
-            suggestion.staff_comments += data["staff_comments"]
-        suggestion.save()
-        return JsonResponse({"result": "success"}, status=200)
-    elif action == "triage_job":
-        if data["progress"] != "JOB_SUB":
-            job: Job = Job.objects.get(pk=data["pk"])
-            job.progress = data["progress"]
-            job.save()
-            if (
-                data["progress"] == "JOB_VFD"
-                and job.payment_preference == "PREF_INVCRD"
-            ):
-                assert job.assignee is not None
-                assert job.semester is not None
-                try:
-                    invoice = Invoice.objects.get(
-                        student__semester=job.semester, student__user=job.assignee.user
-                    )
-                except Invoice.DoesNotExist:
-                    logging.warn(
-                        f"Could not get invoice for {job.assignee.user} and {job.semester}"
-                    )
-                else:
-                    invoice.credits += job.usd_bounty
-                    invoice.save()
-            return JsonResponse({"result": "success", "changed": True}, status=200)
-        else:
-            return JsonResponse({"result": "success", "changed": False}, status=200)
     elif action == "init":
-        output_data: dict[str, Any] = {}
-        output_data["timestamp"] = str(timezone.now())
-        output_data["_name"] = "Root"
+        output_data: dict[str, Any] = {
+            "timestamp": str(timezone.now()),
+            "_name": "Root",
+        }
         output_data["_children"] = [
             {
                 "_name": "Problem sets",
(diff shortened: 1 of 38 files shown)


rpg/models.py
```diff
@@ -156,7 +156,7 @@ class Meta:
         db_table = "dashboard_questcomplete"
 
     def __str__(self) -> str:
-        return self.title + " " + self.timestamp.strftime("%c")
+        return f"{self.title} " + self.timestamp.strftime("%c")
```
Incompleteusern, line 159: might as well fully make this an f-string

rpg/levelsys.py
```diff
@@ -351,24 +351,23 @@ def annotate_student_queryset_with_scores(
...
     max_level = max(levels.keys())
 
     for student in annotate_student_queryset_with_scores(queryset):
-        row: dict[str, Any] = {}
-        row["student"] = student
-        row["spades"] = (getattr(student, "spades_quizzes", 0) or 0) * 2
+        row: dict[str, Any] = {
+            "student": student,
+            "spades": (getattr(student, "spades_quizzes", 0) or 0) * 2,
+        }
```
Incompleteusern, line 370: not really a fan of this one to be honest. I'm also scared why this uses `getattr` 50,000 times.

rpg/levelsys.py
```diff
@@ -351,24 +351,23 @@ def annotate_student_queryset_with_scores(
 
 def compute_insanity_rating(b: int, d: int, z: int) -> float:
     assert min(b, d, z) >= 0
-    if b == 0 and d == 0 and z == 0:
-        return 0
-    return (z - b) / (b + d + z)
+    return 0 if b == 0 and d == 0 and z == 0 else (z - b) / (b + d + z)
```
Incompleteusern, line 354: I think this makes it less clear but I don't know

roster/views.py
```diff
@@ -219,8 +221,12 @@ def master_schedule(request: HttpRequest) -> HttpResponse:
...
         unit_to_student_dicts[d["curriculum"]].append(
             {
-                "short_name": d["user__first_name"] + " " + d["user__last_name"][0:1],
-                "full_name": d["user__first_name"] + " " + d["user__last_name"],
+                "short_name": d["user__first_name"]
+                + " "
+                + d["user__last_name"][:1],
+                "full_name": d["user__first_name"]
+                + " "
+                + d["user__last_name"],
```
Incompleteusern, line 229: this might be better as an f string

roster/utils.py
```diff
@@ -31,10 +31,7 @@ def get_visible_from_queryset(user: User, queryset: QuerySet[models.Student]):
 
 
 def get_visible_students(user: User, current: bool = True):
-    if current:
-        queryset: QuerySet = get_current_students()
-    else:
-        queryset = models.Student.objects.all()
+    queryset = get_current_students() if current else models.Student.objects.all()
```
Incompleteusern, line 34: type check seems to have disappeared here.

Incompleteusern: I also prefer the old formatting but its fine

otisweb/settings.py
```diff
@@ -260,9 +261,9 @@ def filter_useless_404(record: logging.LogRecord) -> bool:
         return True
     a: list[str] = [str(s) for s in record.args]
     if len(a) == 2:
-        return not (a[0] == "Not Found" and ("wp-include" in a[1] or ".php" in a[1]))
+        return a[0] != "Not Found" or "wp-include" not in a[1] and ".php" not in a[1]
     elif len(a) == 3:
-        return not (a[1] == "404" and ("wp-include" in a[0] or ".php" in a[0]))
+        return a[1] != "404" or "wp-include" not in a[0] and ".php" not in a[0]
```
Incompleteusern, line 266: This is less clear IMO.

markets/views.py
```diff
@@ -133,14 +133,16 @@ def dispatch(
...
+            and self.market.has_ended
+        ):
             return super().dispatch(request, *args, **kwargs)
         elif not self.market.has_started:
             return HttpResponseNotFound()
-        elif not self.market.has_ended:
-            return HttpResponseRedirect(self.market.get_absolute_url())
         else:
-            return super().dispatch(request, *args, **kwargs)
+            return HttpResponseRedirect(self.market.get_absolute_url())
```
Incompleteusern, line 145: This deletes `super().dispatch`, is this intentional?

markets/models.py
```diff
@@ -84,14 +84,10 @@ def __str__(self) -> str:
         return f"{self.title} ({self.slug})"
 
     def get_absolute_url(self) -> str:
-        if self.has_ended:
+        if self.has_ended or not self.has_started:
```
Incompleteusern, line 87: IMO flipping the if statement `if self.has_started and not self.has_ended` is clearer

arch/migrations/0013_auto_20210603_2343.py
```diff
@@ -145,12 +145,12 @@
...
-    lookup[k + " RMM TST"] = v + "RST"
-    lookup[k + " EGMO TST"] = v + "EST"
-    lookup[k + " JBMO TST"] = v + "JST"
-    lookup[k + " TSTST"] = v + "TSTST"
+    lookup[f"{k} MO"] = f"{v}MO"
+    lookup[f"{k} TST"] = f"{v}TST"
+    lookup[f"{k} RMM TST"] = f"{v}RST"
+    lookup[f"{k} EGMO TST"] = f"{v}EST"
+    lookup[f"{k} JBMO TST"] = f"{v}JST"
+    lookup[f"{k} TSTST"] = f"{v}TSTST"
```
Incompleteusern, line 153: refactoring a migration seems a bit scary

aincrad/views.py
```diff
@@ -499,7 +499,9 @@ def hanabi_handler(action: str, data: JSONData) -> JsonResponse:
                     name_to_replay_id_dict[name] = replay_json["replay_id"]
 
     for replay_json in data["replays"]:
-        if not any(name in eligible_players for name in replay_json["players"]):
+        if all(
+            name not in eligible_players for name in replay_json["players"]
+        ):
```
Incompleteusern, line 504: Prefer the old one

Incompleteusern (commented):
This probably should've been multiple commits.

arch/migrations/0013_auto_20210603_2343.py
vEnhance: i think it's equivalent, so it should be fine... i hope...

vEnhance:
Thanks for this.

I might wait a bit to merge this on because there's about to be a Django 4.1 -> 4.2 upgrade that might also break a lot of things, so I might want to spread out the massive changes a bit to make it easier to tell where things are breaking.

Unrelatedly, I'm curious whether you made all these changes by hand or whether there was some automation? It's a rather comprehensive PR.

Incompleteusern:
Given that some parts are missed, I think this was done manually.

markets/views.py
vEnhance: think you need to look at all the changed lines
basically it's a DRY on the two super().dispatch cases

vEnhance:
I'll try to merge this sometime over the coming weekend (Aug 19 - Aug 20) to allow a bit of time for the Django 4.2 migration first, just so we don't hit the server with too many changes at once.

vEnhance:
ok it looks like the mysql/django upgrade is in working order so let's merge this now and see if anything breaks :yayy:

vEnhance:
dean I gave you +9 spades as a thank-you for all this cleanup, much appreciated

PR #276, opened 2023-08-15
Test for get_news

### LINKED ISSUE #236: write tests for get_news
need to write this later, but prioritizing getting this onto prod because kickoff is going out tonight and i literally just thought of this feature like 40 minutes ago so hahaha

tbh i bet `dismiss` isn't tested at all right now either smh

secretly hoping someone writes this for me but expecting it to be more like, i eventually force myself to do this once otis stuff dies down a bit

Incompleteusern:
Fixes #236.


vEnhance:
i think `assertTrue` and `assertFalse` are pretty weak checks and it'd be better for there to be length checks, i'll add some now

PR #282, opened 2023-08-24
Fix non_field_errors

### LINKED ISSUE #278: Horrifying red box when Hint Number Already Exists
image

To recreate, make a hint at n% and a hint at n-1%, then edit the form to change n-1% to n%

Incompleteusern:
Fixes #278. Does some hacky things to get non_field_errors to look similar. Issue here was that non_field_errors isn't a dictionary.

image

### DIFF
--- a/otisweb/templates/generic_form.html
+++ b/otisweb/templates/generic_form.html
@@ -8,27 +8,30 @@
     <div class="alert alert-danger" role="alert">
       <strong>{{ "Error"|title }}</strong>
       <ul>
-        {% for error in form.non_field_errors %}<li>{{ error }}{% endfor %}
-      </ul>
-    </div>
-  {% elif form.errors %}
-    <div class="alert alert-danger" role="alert">
-      {% for err,val in form.errors.items %}
-        {% if err == "content" and "This field is required." in val %}
-          <strong>File upload</strong>
-          <ul>
-            <li>
-              There was an error with your file upload.
-              Please make sure that you file submission and is no larger than 10 MB.
-            </li>
-          </ul>
-        {% else %}
-          <strong>{{ err|title }}</strong> {{ val }}
-        {% endif %}
-      {% endfor %}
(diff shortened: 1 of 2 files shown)


otisweb/templates/generic_form.html
```diff
@@ -6,28 +6,32 @@
...
-      {% endfor %}
-    </div>
-  {% endif %}
-  {% for hidden in form.hidden_fields %}{{ hidden }}{% endfor %}
-  {% for field in form.visible_fields %}{{ field|as_crispy_field }}{% endfor %}
-  <button type="submit" class="btn {{ btn_class|default:"btn-success" }}">{{ submit_name|default:"Submit" }}</button>
-</form>
+        {% for error in form.non_field_errors %}
+          <li>
+            {{ error }}
```
vEnhance, line 13: can we add `</li>` here?

otisweb/templates/generic_form.html
```diff
@@ -6,28 +6,32 @@
   {% csrf_token %}
   {% if form.non_field_errors %}
     <div class="alert alert-danger" role="alert">
+      <strong>{{ "Error"|title }}</strong>
```
vEnhance, line 9: why is the `|title` filter necessary here?

vEnhance: maybe better use `<h3 class="alert-heading"></h3>` instead of `<strong>` as well.

vEnhance (changes requested):

otisweb/templates/generic_form.html
vEnhance: this will probably de-dent the part afterwards which is good because it makes the diff smaller

otisweb/templates/generic_form.html
Incompleteusern: It was so it looks similar to the non field error style

PR #289, opened 2023-08-28
Deprecate create_if_missing

### LINKED ISSUE #285: Remove `create_if_missing` for hint list
I've never really seen the reasons to have them differ.

Incompleteusern:
Fixes #285. Merges functionality. The endpoint still exists for backward compatability.


vEnhance:
> The endpoint still exists for backward compatability.

can we actually make it a redirect view to the new URL so that we can phase out the old URL faster

i'll update this in vEnhance/dotfiles and recompile that.

vEnhance:
OK here's the redirect thing, should work now.

PR #293, opened 2023-08-29
upcoming markets/hanabi seeds

### LINKED ISSUE #271: Heads-up on upcoming events
It would be nice if the markets page indicated when the next market was going to
open (by looking in the database for markets that haven't started and taking the
next one), without revealing any details about what would happen in the market.
Builds some hype, you know.

Similar for Hanabi seeds.

waystern:
fixes #271 

current progress

made markets starting within 4 days visible and hanabi seeds starting within a month visible

markets show the slug currently for both student and staff view will change the staff view to be like the others with link and all


markets/models.py
```diff
@@ -101,6 +115,9 @@ def has_started(self) -> bool:
     def has_ended(self) -> bool:
         return timezone.now() >= self.end_date
 
+    @property
+    def is_upcoming(self) -> bool:
+        return timezone.now() < self.start_date and timezone.now()+timedelta(days=7) >= self.start_date
```
vEnhance, line 120: Let's contract this to `timezone.now() < self.start_date < timezone.now() + timedelta(days=7)`, easier to read and shorter

vEnhance (changes requested):
Generally looks really solid. Thanks much.

I left some small feedback. We should also update `hanabicontest_list.html` with the upcoming changes, I think this should be pretty simple, above lines 67-69 add a new case for upcoming contests and hide the variant name/links.

Right now the code doesn't pass the style test. Run `black .` from the root directory of the project and it will autoformat the files so that check passes.

waystern:
@vEnhance sounds good. this was just so you could see my progress not a final product

I’ll make your suggested changes later today when I can get on my laptop

vEnhance (approved):

hanabi/templates/hanabi/hanabicontest_list.html
```diff
@@ -64,6 +64,9 @@ <h1>All Contests</h1>
               <a href="{{ contest.get_absolute_url }}">{{ contest.variant_name }}</a>
             </td>
             <td>Completed</td>
+          {% elif contest.is_upcoming %}
+            <td class="table-primary">Starts in {{ contest.start_date }}</td>
```
vEnhance, line 68: Can we flip the colors so that the upcoming is colored by `warning` and the active is colored by `primary`? (Right now it's vice-versa.)

markets/templates/markets/market_list.html
```diff
@@ -26,10 +26,16 @@
       <th>Ending date</th>
     </tr>
     {% for market in markets %}
-      <tr class="{% if not market.has_started %}table-warning{% elif not market.has_ended %}table-success{% endif %}">
-        <td>
-          <a href="{{ market.get_absolute_url }}">{{ market.title }}</a>
-        </td>
+      <tr class="{% if market.is_upcoming %}table-primary{% elif not market.has_started %}table-warning{% elif not market.has_ended %}table-success{% endif %}">
```
vEnhance, line 29: Again I think colors should be flipped here; upcoming is better as yellow imo

maybe make this consistent with the markets template too

vEnhance (changes requested):
Great!

Run `djlint` on the templates (or just run `autofix.sh`) to pass that style check, and change a couple of the colors to be consistent between the templates (maybe we can do warning/yellow for upcoming, success/green for active, primary/blue for past as the standard).

vEnhance (approved):

PR #321, opened 2023-10-26
Decouple Difficulty and Version Fields

### LINKED ISSUE #317: Split Unit code into properties
Currently the `Unit` model uses
```python
    code = models.CharField(
        max_length=255,
        help_text="The version code for the handout, like 'ZGX'",
    )
```
which is a bit of a hardcode, especially as subject is already specified by its unit group.
Splitting this field into a version choice field and a difficulty choice field might be better, and constructing the code can be done as a property.

Incompleteusern:
Fixes #317.

Implements version codes as
```python
DIFFICULTY_CHOICES = (
    ("B", "Easy/Bet"),
    ("D", "Medium/Dalet"),
    ("Z", "Hard/Zayin"),
)

VERSION_CHOICES = (
    ("V", "V (Secret)"),
    ("W", "W"),
    ("X", "X"),
    ("Y", "Y"),
)

difficulty = models.CharField(
    max_length=2,
    choices=DIFFICULTY_CHOICES,
    help_text="The difficulty code for the handout, like B",
)
version = models.CharField(
    max_length=2,
    choices=VERSION_CHOICES,
    help_text="The version code for the handout, like W",
)
```

Benefits of this include being less prone to typos and less redundant information stored as the group subject.

Other things this implies
1. Migrate the database. Probably make a backup before that? See the custom code in the migration.
2. There's likely some latent bugs here. Will fix as notified. Free 500 diamond I suppose.
3. venueQ needs to be updated for this
4. fixtures should be re-dumped to conform to this format. Maybe deprecate all.json, though not sure anymore.


core/templates/core/unit_list.html
```diff
@@ -19,7 +19,8 @@
     <li>The second letter repeats the olympiad category.</li>
     <li>
       The third letter is a version identifier (either W, X, or Y).
-      Many topics have multiple versions so they can be repeated in different years.
+      Many topics have multiple versions so they can be repeated in different years,
+      but you can do the same topic multiple times in one year as well.
```
Incompleteusern, line 23: Added this comment since I misinterpreted it the first time.

rpg/tests.py
```diff
@@ -115,7 +115,7 @@ def test_level_up(self):
         alice = self.get_alice()
         self.login(alice)
         bonus = BonusLevelFactory.create(group__name="Level 40 Quest", level=40)
-        bonus_unit = UnitFactory.create(group=bonus.group, code="DKU")
```
Incompleteusern, line 118: Cursed U version code?

jatloe: WHAT IS THIS

jatloe: what's "Level 40 Quest"

Incompleteusern: It refers to a secret unit level.

jatloe: i prefer meow :3

core/migrations/0058_alter_unit_unique_together_unit_difficulty_and_more.py
```diff
@@ -0,0 +1,57 @@
+# Generated by Django 4.2.5 on 2023-10-26 01:26
+
+from django.db import migrations, models
+
+
+def set_difficulty_and_version(apps, schema_editor):
+    # We can't import the Person model directly as it may be a newer
```
vEnhance, line 7: seems to be a copy paste error?

core/migrations/0058_alter_unit_unique_together_unit_difficulty_and_more.py
```diff
@@ -0,0 +1,57 @@
...
+            name="version",
+            field=models.CharField(
+                choices=[("W", "W"), ("X", "X"), ("Y", "Y")],
+                default="W",
+                help_text="The version code for the handout, like W",
+                max_length=2,
+            ),
+            preserve_default=False,
+        ),
+        migrations.RunPython(set_difficulty_and_version),
```
vEnhance, line 48: doesn't RunPython need a reverse operation argument or something like that? at least i feel like i've had to specified noop a bunch before in the past

Incompleteusern: I think that its optional? Stackoverflow had that at some point and it seems to migrate properly.

vEnhance:
although this is probably the way the code should have been written, i'm a bit hesitant to merge this at a time of year where i expect to be pretty busy. this is a pretty big migration, a lot of code changes (i feel like i'd need to test this a lot before i feel like it's safe to merge) and it would require venueQ to be updated as well

so i'll leave this open for now, but it'll be a while before i get to actually processing it

vEnhance:
okay i've thought about this more and i've decided that fixing this is probably not worth it, due to all the changes that would be needed in many places (in particular, i'm sure we won't catch everything, even with the unit tests). "there's likely some latent bugs here" is a worse outcome to me than a small amount of extra redundancy in the database.

Incompleteusern:
Yeah that makes sense I suppose.

Issue #318, opened 2023-10-22
Allow edits to previous years unit submissions


anterick-cjx:
Basically I tried to "submit additional problems" for last year's unit since otherwise I'd have to request it again and that takes too long, but it says you can't do that because the semester isn't active. I think it'd be a lot more efficient if we were allowed to submit from previous years or something like that.

Incompleteusern:
I think the consensus is that https://github.com/vEnhance/otis-web/issues/300 is the easiest way of implementing this.

vEnhance:
I agree that finishing #300 (i.e. allowing automatic approval of requests of units that were unlocked in a previous year) would provide a workaround here that isn't too annoying.

It's more complicated to try and resubmit to an inactive semester, because the pset/unit logic is based on ordered pairs of (human, year) rather than just the human. So, e.g. if a student submits a unit in year Y and then resubmits it in year Y+1 it's no longer clear which year that submission should be attached to.

In the long term maybe we should just drop the year dependence; https://github.com/vEnhance/otis-web/commit/1ebab2a6 was a related issue caused by the extra dependence. However, that's quite long-term because it requires refactoring a ton of code.

PR #322, opened 2023-10-26
Auto accept previous year units

### LINKED ISSUE #300: Auto-Unlock previous year units
> Units a person had in previous years could just be automatically unlocked in unit petition. That might be the cleanest way.

Incompleteusern:
Fixes #300 #318.

Automatically accepts petitions of a user that has students of another instance with that unit unlocked.


vEnhance:
Some changes requested:

1. There's some random print statements that need to be fixed
2. I think this is an O(n) query where n is the number of times you've done OTIS, but it really should be O(1). Django should support something like `Student.objects.filter(user=student.user, curriculum__<something>=unit)`, I don't know off my head but I'd be shocked if it didn't exist.
3. Needs unit tests.

vEnhance:
still waiting on unit tests here

jatloe:
is this dead

Incompleteusern:
I don't feel like working on this right now. I might finish unit tests when this OTIS semester is over.

palindrome868wastaken:
what are unit tests?

Incompleteusern:
Unit Tests should be done.

chessisdabest:
Wonderful timing for the new year...

PR #342, opened 2024-04-07
Fixes #326

### LINKED ISSUE #326: Doc: Explain emojis at /core/catalog
We should list that

- check mark means done
- blue clock means unlocked but not done
- grey hourglass means locked but on to-do list
- bold black is everything else

owennaonao:
Added catalog description. - Owen Zhou

(I think I did everything correctly??? Idk first time using GitHub)


vEnhance:
1. You should not commit .DS_Store` files, those are [garbage](https://en.wikipedia.org/wiki/.DS_Store). If you have a global gitignore it'd be good to include that actually.
2. Include the emoji itself in addition to just the text, e.g. "A check mark (✔)"
3. Try to have the formatting pass djlint.

owennaonao:
Um does the above change fix the first two points? I'm not sure what's wrong with the formatting, but I just legit copy pasted the formatting originally there and changed the text.

Also, when I pushed the second commit to GitHub, it just instantly popped up here. Is there a way to have it only show up on my forked repository? Idk if I'm explaining this well.

Incompleteusern:
> Also, when I pushed the second commit to GitHub, it just instantly popped up here. Is there a way to have it only show up on my forked repository? Idk if I'm explaining this well.

No because the merge request comparison is done by comparing forks.

To do the formatting, run djlint locally (or the equivalent bash script that does all that)

owennaonao:
Could someone check where in the world my style is failing? I've never coded hmtl before, so I just copied the previous style, but apparently something's wrong??? Other than that, the first two points Evan mentioned should be fixed?

vEnhance:
> Could someone check where in the world my style is failing? I've never coded hmtl before, so I just copied the previous style, but apparently something's wrong??? Other than that, the first two points Evan mentioned should be fixed?

Run
`djlint --reformat core/templates/core/unit_list.html`
and it will fix the formatting. (Then commit and push the changes.)

owennaonao:
Alright fixed it.

Issue #332, opened 2024-01-02
Press Esc to exit a hint on ARCH


jatloe:
Should have the same effect as "Back to ARCHCODE"

MrGliff:
I can do it!

jatloe:
@MrGliff how is this going?

palindrome868wastaken:
Starting on it. ig just some js

palindrome868wastaken:
Ok. Now I have time to retry this.

Kempu33334:
I think that this can be closed now.

jatloe:
im probably missing something but this just doesnt work?

vEnhance:
i didn't deploy to prod yet i think let me do that now

PR #349, opened 2024-05-17
Escape Key to Exit Hint on ARCH

palindrome868wastaken:
#332 
e.g. enhancement or feature update, bug fix, typos or copy edits, etc.

Enhancement feature update.
Passes python manage.py check.

Username: ducky7629

### DIFF
--- a/sync_static
+++ b/sync_static
@@ -2,7 +2,4 @@
 
 # If you aren't Evan, replace the bucket name...
 python manage.py collectstatic --noinput
-gsutil -m rsync -R static/ gs://otisweb-static/static
-
-# gsutil cors set cors.json gs://otisweb-static/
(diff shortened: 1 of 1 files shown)


vEnhance:
Few things:

1. You should use a four-space indent for the JavaScript for consistency with rest of project, not a three-space indent. (If you run ./autofix.sh and you installed everything properly, it will autofix this for you. In general, good projects will have tools automatically fixing this kind of style thing so you don't have to do it by hand.)
2. Can you add a bit of text somewhere saying "Press ESCAPE to return" or whatever on the page so that people know this feature exists? Ideally, show a screenshot of what it looks like too so I can see it.

Once you've done those, first test it locally to make sure it does what you want, and then commit and push the new changes.

Thanks!

jatloe:
wao

jatloe:
Might want to check that clicking the exit hint and pressing escape, or pressing escape on a different screen, doesn't have any negative side effects

PR #364, opened 2024-11-09
Catalog qol

zvfrozel:
pretty comprehensive(?) filtering and sorting functionality in the unit catalog to enhance Quality of Life (ft. ChatGPT)

Code passes style checks but tests haven't been written.

image


core/templates/core/unit_search_form.html
```diff
@@ -0,0 +1,86 @@
...
+        <button type="submit" class="btn btn-primary d-inline-block">Apply Filters</button>
+      </div>
+    </div>
+  </form>
+</div>
+{% if not units.exists %}
+  <div class="alert alert-warning mt-3" role="alert">
+    No units found matching your search criteria. 👿
+    <br>
+    ae0de5c24b40dd370cfdf64das
```
Incompleteusern, line 69: What is this supposed to be?

vEnhance:
- The hints_released_date thing should really be a separate PR, not merged into, but anyway I added it in myself to `main` so that's a moot point.
- It looks like one of your commits seems to have accidentally broken the symlink to `otisweb/templates/socialaccount/base.html`. I'll revert this for you and see if it passes tests after making that change.

More comments forthcoming once I actually read the PR and try this on my local server.

core/templates/core/unit_search_form.html
zvfrozel: Admits

vEnhance:
This looks really good! I made a few cosmetic changes to personal taste, but overall I think this looks really solid.

If you go ahead and write a few unit tests we can basically merge it in right away since I've checked that it seems to work the way intended.

zvfrozel:
Thanks 🧡
I (chatgpt) will write the unit tests shortly and update the PR, no problem

zvfrozel:
Ok, i think we are good to go.

Is it possible to disable linting for specific files? case in point - ruff made the test_catalog_filters function in tests.py look downright ugly.

vEnhance:
> Is it possible to disable linting for specific files?

yes but it's not worth it i think

vEnhance:
ok i looked at the commit a981959 and in general if you want ruff to put every argument on its own line you should put a trailing comma, and it'll do so, se 1915baf for an example of how to do this

vEnhance:
alright i think this is good now thanks much, it should be on production momentarily.

thank you for this much-needed pull request!

zvfrozel:
there is a typo here by the way:
image

vEnhance:
OK, but in the future feel free to just open a pull request yourself if you have patches like this you want to merge.

zvfrozel:
Okay

PR #367, opened 2024-11-16
Catalog patch - fix 500 error

zvfrozel:
Fix the 500 error when an inactive user tries to filter the catalog.

Additionally, require a login to access the (private) catalog.

### DIFF
--- a/core/forms.py
+++ b/core/forms.py
@@ -3,7 +3,7 @@ from django import forms
 from .models import UnitGroup
 
 
-class FilterForm(forms.Form):
+class CatalogFilterForm(forms.Form):
     difficulty = forms.MultipleChoiceField(
         choices=[("B", "B"), ("D", "D"), ("Z", "Z")],
         widget=forms.CheckboxSelectMultiple(attrs={"class": "filter-form"}),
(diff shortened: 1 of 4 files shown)


vEnhance:
i don't know what text editor you're using, but whatever it is, you managed to convert the file into DOS encoding, which is why in `core/views.py` if you look at the diff it shows that every single line has changed, which is really bad

vEnhance:
look at your commit 4cee8364 for an example of what not to do (note that every line is marked as changed, because your text editor thought it was a good idea to modify the ending on every single line in the file from the standard LF to CRLF).

zvfrozel:
Vscode.
Initially the commit had rendered as if every single file in the repo was changed, I somehow git rebased to end up like this

vEnhance:
> Initially the commit had rendered as if every single file in the repo was changed, I somehow git rebased to end up like this

sounds on brand for windows.

considrer running `git config --global core.autocrlf input` to fix your setup (well, maybe VSCode is the issue).

zvfrozel:
> 
> sounds on brand for windows.

Issue #210, opened 2023-04-24
Show Unlocked/Finished Secret Units in Catalog


Incompleteusern:
This might be possible.

vEnhance:
It should be possible, I just don't know if it's worth the extra code complexity.

zvfrozel:
Fixed at https://github.com/vEnhance/otis-web/pull/373

PR #373, opened 2024-12-21
Unlocked secret units in catalog

zvfrozel:
Secret units don't show up in the catalog even if they're unlocked, this fixes that.

Fixes https://github.com/vEnhance/otis-web/issues/210

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -56,8 +56,7 @@ class UnitGroupListView(LoginRequiredMixin, ListView[Unit]):
 
         if student:
             queryset = Unit.objects.exclude(
-                group__hidden=True,
-                group__bonuslevel__level__gt=student.last_level_seen
+                group__hidden=True, group__bonuslevel__level__gt=student.last_level_seen
             )
         elif user.is_staff:
             queryset = Unit.objects.all()
--- a/core/views.py
+++ b/core/views.py
@@ -46,7 +46,14 @@ class UnitGroupListView(LoginRequiredMixin, ListView[Unit]):
     template_name = "core/unit_list.html"
 
     def get_queryset(self):
-        if student := Student.objects.filter(user=self.request.user).first():
+        student = Student.objects.filter(user=self.request.user, semester__active=True).first()
+        if student is None:
+            student = Student.objects.filter(user=self.request.user).order_by("-pk").first()
+            active = False
+        else:
+            active = True
+
+        if student:
             queryset = Unit.objects.exclude(
                 group__hidden=True,
                 group__bonuslevel__level__gt=student.last_level_seen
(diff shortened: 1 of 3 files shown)


core/views.py
```diff
@@ -46,7 +46,22 @@ class UnitGroupListView(LoginRequiredMixin, ListView[Unit]):
...
+            student = Student.objects.filter(user=user).order_by("-pk").first()
+            active = False
+        else:
+            active = True
+
+        if student:
+            queryset = Unit.objects.exclude(
+                group__hidden=True, group__bonuslevel__level__gt=student.last_level_seen
+            )
+        elif user.is_staff:
```
vEnhance, line 61: The staff check should come before the student check I think. (Most staff are OTIS alum's!)

core/views.py
```diff
@@ -46,7 +46,22 @@ class UnitGroupListView(LoginRequiredMixin, ListView[Unit]):
     template_name = "core/unit_list.html"
 
     def get_queryset(self):
-        queryset = Unit.objects.filter(group__hidden=False)
+        assert isinstance(user := self.request.user, User)
+        student = Student.objects.filter(user=user, semester__active=True).first()
```
vEnhance, line 50: I think this has the effect that, e.g. once summer starts, the secret units disappear, and only reappear if the student re-enrolls in OTIS.

I'd suggest that we actually take the `Max` of `last_level_seen` across all the students? (Of course then we should handle the case of empty `Max`.)

vEnhance: Something like the following (please test and make sure it actually works :P )

```python
level : int | None = Student.objects.filter(user=user).aggregate(
    level=Max("last_level_seen")
)["level"]
```

vEnhance (changes requested):
Looks pretty good, two minor suggestions on logic changes I'd like to see if you agree with, then we can merge this.

core/views.py
zvfrozel: This is correct, gone and done that

zvfrozel:
Should be fixed.

> I think this has the effect that, e.g. once summer starts, the secret units disappear, and only reappear if the student re-enrolls in OTIS.

i think the pk logic should handle that unless a new student instance for next semster is created over the summer.
But I changed the code to the Max aggregate logic anyway.

vEnhance:
> i think the pk logic should handle that unless a new student instance for next semster is created over the summer.

Hmm, I don't see how. 

> test_hidden fails now!

Fix the code! The test failing is telling you that something is wrong. If we just ignore it by deleting the test, that defeats the purpose of having the test.

In fact this PR should _add_ new tests to check the logic for these secret units.

zvfrozel:
> In fact this PR should add new tests to check the logic for these secret units.

Yes, of course it should! I've fixed the broken tests now and added more tests for the new logic

I had left the PR incomplete halfway through yesterday, sorry about that.

vEnhance (approved):

vEnhance:
> I had left the PR incomplete halfway through yesterday, sorry about that.

ah, okay, mkaes sense. No worries!

This looks great now, thanks for making it happen.

PR #376, opened 2025-01-12
fix(opal): Change title of rules page

zvfrozel:
The title says Author instructions


vEnhance:
Thanks! Looks like ruff has a minor version update so I'll deal with that on my end.

PR #383, opened 2025-04-28
edit: Added escape feature to ARCH

Kempu33334:
When clicking escape from an ARCH hints page, it should lead back to the ARCH homepage.

This is intended to fix [Issue #332](https://github.com/vEnhance/otis-web/issues/332).

OTIS-WEB username: Kempu33334
ID number: 2312

If there is still a problem, please let me know.

### DIFF
--- a/arch/templates/arch/hint_list.html
+++ b/arch/templates/arch/hint_list.html
@@ -5,7 +5,7 @@
 {% endblock title %}
 {% block layout-content %}
   <p>
-    <a href="{% url "arch-index" %}"><span class="emoji-text">⬅️</span> Back to ARCH index</a>
+    <a href="{% url "arch-index" %}" id="backToArch"><span class="emoji-text">⬅️</span> Back to ARCH index</a>
   </p>
   <hr />
   {% spaceless %}
@@ -75,3 +75,15 @@
     </a>
   </div>
 {% endblock layout-content %}
+{% block scripts %}
+<script defer>
+    document.addEventListener('keydown', function (event) {
+        if (event.key === 'Escape') {
+            console.log('Escape pressed');
+            const link = document.getElementById('backToArch');
+            if (link) {
+                link.click();
+            }
+        }
+    });
+</script>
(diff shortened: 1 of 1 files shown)


arch/templates/arch/hint_list.html
```diff
@@ -75,3 +75,15 @@ <h2>Hint {{ hint.number }}%</h2>
     </a>
   </div>
 {% endblock layout-content %}
+
+<script defer>
+    document.addEventListener('keydown', function (event) {
+        if (event.key === 'Escape') {
+            console.log('Escape pressed');
```
vEnhance, line 82: Please remove debug statements like this.

arch/templates/arch/hint_list.html
```diff
@@ -75,3 +75,15 @@ <h2>Hint {{ hint.number }}%</h2>
...
+            const link = document.getElementById('backToArch');
```
vEnhance, line 83: Finding the link element by `document.getElementById` seems like a bit of a hack, I would just replace the location directly rather than clicking a link. You can for example set `window.location.href` or something like this.

arch/templates/arch/hint_list.html
```diff
@@ -75,3 +75,15 @@ <h2>Hint {{ hint.number }}%</h2>
     </a>
   </div>
 {% endblock layout-content %}
+
```
vEnhance, line 78: Scripts need to be enclosed in `{% block scripts %}` ... `{% endblock scripts %}` to show up.

vEnhance (changes requested):
- Please test your changes before submitting a PR, you should see README for how to spin up a local server. Unless I"m missing something, this doesn't seem to work when I test it locally.
- I think you are misunderstanding the issue. This PR tries to add functionality for escape going from a list of hints for a problem like USAMO 20XX/Y back to the homepage. But the issue asks for going from a specific hint (like 10% for USAMO 20XX/Y) to the list of hints for USAMO 20XX/Y.

PR #386, opened 2025-05-13
polish: Added word wrapping for long urls, better UI/visual quality on small screens.

Kempu33334:
I added a little bit of css to make the long text wrap.

Please let me know if (when) there are problems.

Username: Kempu33334
ID: 2312

### DIFF
--- a/arch/templates/arch/hint_detail.html
+++ b/arch/templates/arch/hint_detail.html
@@ -6,7 +6,7 @@
 {% endblock title %}
 {% block layout-content %}
   <p>Here is your hint:</p>
-  <div class="alert alert-info" style="word-wrap: break-word;">{{ hint.content|parbreaks }}</div>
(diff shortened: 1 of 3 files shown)


vEnhance:
Thanks, let's give this a spin. I agree using this on diamond code makes sense, I'm a bit more nervous about hint text, but we can try it.

Incidentally, I realized there is a bootstrap class that handles this. That's probably better than doing inline style, so I switched to that class.

Kempu33334:
_“… I'm a bit more nervous about hint text …”_

I actually didn’t see any active need for it, but added it just in case (don’t think it can do any harm) and because I felt it was a similar situation to the hints page.

vEnhance:
> don’t think it can do any harm

if it breaks actual words over an end of line i would argue that's not good, e.g. if "take the reciprocal of x" becomes

```
take the reci
procal of x
```

so I think I'll leave it off the hint text itself for now. I agree it makes sense for long URL or diamond code

PR #389, opened 2025-05-22
feat: Update hint_detail.html

Kempu33334:
Solves https://github.com/vEnhance/otis-web/issues/332.

I added JavaScript code to the hint_detail.html so that when the escape key is pressed, it navigates back to the list of hints page.

(It works on my end, but please let me know if it doesn't work so I can fix it.)

Username: Kempu33334
ID: 2312

### DIFF
--- a/arch/templates/arch/hint_detail.html
+++ b/arch/templates/arch/hint_detail.html
@@ -28,10 +28,10 @@
 {% endblock layout-content %}
 {% block scripts %}
   <script type="text/javascript">
-      document.addEventListener('keydown', function (event) {
-        if (event.key === 'Escape' || event.key === 'Esc') {
-          window.parent.location.href = document.referrer;
-        }
+      document.addEventListener('keydown', function(event) {
+          if (event.key === 'Escape' || event.key === 'Esc') {
+              window.parent.location.href = document.referrer;
+          }
       });
   </script>
 {% endblock scripts %}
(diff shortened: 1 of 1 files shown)


arch/templates/arch/hint_detail.html
```diff
@@ -26,3 +26,12 @@
     </li>
   </ul>
 {% endblock layout-content %}
+{% block scripts %}
+  <script type="text/javascript">
+      document.addEventListener('keydown', function(event) {
+          if (event.key === 'Escape' || event.key === 'Esc') {
+              window.parent.location.href = document.referrer;
```
vEnhance, line 33: I don't think we should use `document.referrer` here; for example, someone might navigate to the hint page directly from a link given by the Discord bot in the OTIS Discord server, in which case there is no referrer.

Instead we should fill in the URL to the problem using `{% url %}`.

vEnhance (changes requested):

arch/templates/arch/hint_detail.html
vEnhance: I think it'd be better to do it with Django. If you're up to try it, the point is that you should replace the RHS with something like `{% url "hint-list" problem.puid %}` which the template engine will then replace (look at line 8 of `vote_form.html` for an eaxmple; it doesn't matter that it's in a JS block instead of an href).

If you're up to try it go for it; otherwise I'll stick it in myself later after my graduation ceremony later this week.

Kempu33334: I think I want to try it (just because I want to get a little familiar and experiment with Django).

vEnhance (approved):

Issue #378, opened 2025-02-13
OPAL partial answer confirmations


vEnhance:
Certain cluephrases might be better as "Keep going!" or similar rather than just being marked as wrong, e.g. for the gazillion people that didn't realize "CALL IN X" means to call in X as the answer.

vEnhance:
this needs to be done in the next couple months for the impending OPAL 2. it's not difficult so i can do it myself later in summer, but if anyone wants to jump on it while i'm preoccupied with IMO coordination, go for it!

Kempu33334:
Any ideas for what clue-phrases to add?

Here's what I have currently:
Let X be the answer to the puzzle and Y be any string (possibly X).

- CALL IN Y
- PUT IN Y
- X+"s"

PR #397, opened 2025-07-03
feat: Added OPAL partial answer confirmation, issue #378

Kempu33334:
Added in OPAL partial answer confirmation. I'm not exactly sure if it works completely, so would appreciate another set of eyes.

**Note: I also fixed the diamond character counter.** if you submitted a diamond (whether correct or not), the character counter would go down to zero/be re-initialized even though there would still be text in the form (the previous diamond code).

_Username: Kempu33334
ID: 2312_


Kempu33334:
> We need to then factor that into the rate limit (i.e. it does not count against the 20 guess limit)

It was my impression that it's still going to count as an attempt, but the user will get a note that the answer was close --- should we make it so that it doesn't count?

> Change the way the answer is displayed in the history

Does this look ok?
image

vEnhance:
> It was my impression that it's still going to count as an attempt, but the user will get a note that the answer was close --- should we make it so that it doesn't count?

I think so.

> Does this look ok?

I'd prefer "Keep going..." in black

Kempu33334:
Ok, should be done now. Could someone look it over one time, and sanity-check it?

vEnhance:
I will take a closer look when I am back home after IMO, but a couple things right now:
- I would rather have a newline separated list of partial answers than a comma separated one, should be easy change
- I don't think there is a need for a field `excused` (and it does not show up in the migration right now either) which right now literally just copies the value. I think you should just check the number of incorrect answers by filtering on `is_correct=False` and `is_close=False`.
- The counter seems to only be altered in frontend. There should be something in backend updating the logic as well.
- There should be some unit tests to make sure the behavior works as intended.

vEnhance (changes requested):
ok i take back some comments i made --- i realized i have no brain and `excused` was a field i created 11 months ago. sorry, i'm stupid.

but i think we should leave `excused` for manual admin override, rather than tampering with it when is_close=True.

Kempu33334:
Ok, I think I've fixed most of the issues. Although, I'm not exactly sure what this means:

#There should be some unit tests to make sure the behavior works as intended.

vEnhance:
Much thanks, this looks like it works well.

I'm going to compress the migrations here into just a single one, so for your own dev environment, you should rollback to 0012 to get your version to match up, i.e. on your local instance run

```
python manage.py migrate opal 0012
python manage.py migrate
```

or you can just recreate the database, will work too.

vEnhance:
awarded 9 spades

PR #401, opened 2025-07-21
feat: Allow cancel petitions (closes #344)

### LINKED ISSUE #344: Ability to cancel unit petition while pending.
When unlock now/drop unit petition pending for Evan, allow it to be cancelled.

Kempu33334:
added in #344 on petition canceling

_Username: Kempu33334
ID: 2312_

### DIFF
--- a/roster/tests.py
+++ b/roster/tests.py
@@ -1076,7 +1076,7 @@ class RosterTest(EvanTestCase):
     def test_only_owner_or_staff_can_cancel(self):
         alice = StudentFactory.create()
         bob = StudentFactory.create()
-        staff = UserFactory.create(is_staff=True)
+        staff = UserFactory.create(is_staff=True, is_superuser=True)
         unit = UnitFactory.create()
         inquiry = UnitInquiry.objects.create(
             student=alice,
@@ -1087,11 +1087,13 @@ class RosterTest(EvanTestCase):
         )
         # Bob cannot cancel Alice's inquiry
         self.login(bob)
-        resp = self.post("inquiry-cancel", inquiry.pk, follow=True)
-        self.assertEqual(resp.status_code, 403)
+        self.assertPost40X("inquiry-cancel", inquiry.pk, follow=True)
+        inquiry.refresh_from_db()
+        self.assertEqual(inquiry.status, "INQ_NEW")  # Ensure status is still "INQ_NEW"
+
         # Staff can cancel
         self.login(staff)
-        resp = self.post("inquiry-cancel", inquiry.pk, follow=True)
+        self.assertPost20X("inquiry-cancel", inquiry.pk, follow=True)
         inquiry.refresh_from_db()
         self.assertEqual(inquiry.status, "INQ_CANC")
 
@@ -1107,8 +1109,7 @@ class RosterTest(EvanTestCase):
                 explanation="Test",
             )
             self.login(alice)
-            resp = self.get("inquiry", alice.pk)
-            self.assertNotIn("Cancel", resp.content.decode())
+            self.assertGet20X("inquiry", alice.pk)
 
     def test_cannot_cancel_non_pending(self):
         alice = StudentFactory.create()
--- a/roster/tests.py
+++ b/roster/tests.py
@@ -1052,3 +1052,76 @@ class RosterTest(EvanTestCase):
         self.assertEqual(alice.assistant.pk, assistant.pk)
         resp = self.assertGetOK("link-assistant")
         self.assertEqual(len(resp.context["form"].fields["student"].queryset), 1)
+
+    def test_cancel_inquiry_sets_status_to_canceled(self):
+        alice = StudentFactory.create()
+        unit = UnitFactory.create()
+        inquiry = UnitInquiry.objects.create(
+            student=alice,
+            unit=unit,
+            action_type="INQ_ACT_UNLOCK",
+            status="INQ_NEW",
+            explanation="Please unlock",
+        )
+        self.login(alice)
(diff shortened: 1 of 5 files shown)


roster/views.py
```diff
@@ -409,6 +409,18 @@ def inquiry(request: AuthHttpRequest, student_pk: int) -> HttpResponse:
     return render(request, "roster/inquiry.html", context)
 
 
+@login_required
+@require_POST
+def cancel_inquiry(request: AuthHttpRequest, pk: int) -> HttpResponse:
+    inquiry = get_object_or_404(UnitInquiry, pk=pk)
+    if inquiry.student.user != request.user and not request.user.is_staff:
+        raise PermissionDenied("You are not authorized to cancel this inquiry.")
+    inquiry.status = "INQ_CANC"
```
vEnhance, line 418: you need to modify the inquiry model in order for this to actually work, `status` is not just an arbitrary string

vEnhance (changes requested):

roster/views.py
```diff
@@ -409,6 +409,20 @@ def inquiry(request: AuthHttpRequest, student_pk: int) -> HttpResponse:
...
 
 
+@login_required
+@require_POST
+def cancel_inquiry(request: AuthHttpRequest, pk: int) -> HttpResponse:
+    inquiry = get_object_or_404(UnitInquiry, pk=pk)
+    if inquiry.student.user != request.user and not request.user.is_staff:
+        raise PermissionDenied("You are not authorized to cancel this inquiry.")
+    if inquiry.status != "INQ_NEW":
+        return HttpResponse(status=403)
```
vEnhance, line 419: i think we should also raise PermissionDenied (or SuspiciousOperation for HTTP 401 can work too).

roster/tests.py
```diff
@@ -1052,3 +1052,76 @@ def test_link_assistant(self) -> None:
...
+        for status in ["INQ_ACC", "INQ_REJ", "INQ_HOLD", "INQ_CANC"]:
+            inquiry = UnitInquiry.objects.create(
+                student=alice,
+                unit=unit,
+                action_type="INQ_ACT_UNLOCK",
+                status=status,
+                explanation="Test",
+            )
+            self.login(alice)
+            self.post("inquiry-cancel", inquiry.pk, follow=True)
```
vEnhance, line 1125: we should also check the status code is 40x

roster/tests.py
```diff
@@ -1052,3 +1052,76 @@ def test_link_assistant(self) -> None:
...
+        unit = UnitFactory.create()
+        inquiry = UnitInquiry.objects.create(
+            student=alice,
+            unit=unit,
+            action_type="INQ_ACT_UNLOCK",
+            status="INQ_NEW",
+            explanation="Please unlock",
+        )
+        self.login(alice)
+        resp = self.post(
```
vEnhance, line 1067: should check the status code is 200

vEnhance: from `evan_django_tools` there's a shortcut for just `self.assertPost20X` replacing `self.post` for this

roster/tests.py
vEnhance: from evan_django_tools there's a shortcut for just `self.assertPost40X` replacing `self.post` for this

althouh really, for a 40X response you probably don't need to inspect the content, the status code is enough in this case imo

vEnhance (changes requested):
looks nice, a couple minor revisions requested and then we can merge this

roster/views.py
Kempu33334: I replaced it with PermissionDenied (instead of adding another raise after the return which wouldn't do anything).

roster/tests.py
```diff
@@ -1052,3 +1052,76 @@ def test_link_assistant(self) -> None:
...
+        inquiry = UnitInquiry.objects.create(
+            student=alice,
+            unit=unit,
+            action_type="INQ_ACT_UNLOCK",
+            status="INQ_NEW",
+            explanation="Please unlock",
+        )
+        # Bob cannot cancel Alice's inquiry
+        self.login(bob)
+        resp = self.post("inquiry-cancel", inquiry.pk, follow=True)
```
vEnhance, line 1090: change this to self.assertPost40X

roster/tests.py
```diff
@@ -1052,3 +1052,76 @@ def test_link_assistant(self) -> None:
...
+        self.assertEqual(resp.status_code, 403)
+        # Staff can cancel
+        self.login(staff)
+        resp = self.post("inquiry-cancel", inquiry.pk, follow=True)
```
vEnhance, line 1094: change this to self.assertPost20X

roster/tests.py
```diff
@@ -1052,3 +1052,76 @@ def test_link_assistant(self) -> None:
...
+            unit = UnitFactory.create()
+            UnitInquiry.objects.create(
+                student=alice,
+                unit=unit,
+                action_type="INQ_ACT_UNLOCK",
+                status=status,
+                explanation="Test",
+            )
+            self.login(alice)
+            resp = self.get("inquiry", alice.pk)
```
vEnhance, line 1110: change this to self.assertGet20X

vEnhance (changes requested):
still a few more self.post and self.get that should have the status code check baked in

basically the point is that
```python
resp = self.post(...)
self.assertEqual(resp.status_code, 403)
```
should just be
```python
self.assertPost40X(...)
```
and other requests that should give 20x status codes should just use `self.assertGet20X` to add one extra check that it's not a 300/400 return code on the request.

Issue #311, opened 2023-09-29
Ability to repeatedly change market submissions.


Incompleteusern:
It'd be funny, I think.

jatloe:
yes i need this

jatloe:
it's like almost a qol feature instead of submitting at the last minute

PR #402, opened 2025-07-24
feat: Added issue #311 on changing market guess

Kempu33334:
Added capability to change guesses to a market. <h6>_this might be a bad idea_</h6>

_Username: Kempu33334
ID: 2312_


markets/models.py
```diff
@@ -96,10 +96,7 @@ def __str__(self) -> str:
         return f"{self.title} ({self.slug})"
 
     def get_absolute_url(self) -> str:
-        if self.has_ended or not self.has_started:
```
vEnhance, line 99: actually now that i'm looking at this i don't see why this was changed either.

markets/models.py
```diff
@@ -166,7 +163,7 @@ def __str__(self) -> str:
         return f"Guessed {self.value} at {self.created_at}"
 
     def get_absolute_url(self) -> str:
-        return reverse("market-results", args=(self.market.slug,))
```
vEnhance, line 169: i'm not sure this should be changed either tbh, although i could understand if it pointed to the updateview

Kempu33334:
Yeah, those are good points. I think the amount of code just isn't worth it right now --- I'll come back to this when I get time, but I think I'll leave it alone for now.

Issue #273, opened 2023-08-14
Recover old notifications / notifications log


jatloe:
(me when im the kind of person to leave a notification on for a week before closing it)

jatloe:
![image](https://github.com/vEnhance/otis-web/assets/85522239/c94b9dce-f458-4691-835e-7716fa6eb685)
im talking about this btw

Incompleteusern:
What does this mean?

jatloe:
Like some tab to show every "news" thing in the past

Incompleteusern:
Most news items stay you dismiss them

Kempu33334:
Could I "repurpose" the announcement button/icon? Instead of redirecting to MailChimp, it would direct to the list of news things, and there would be a button to go to MailChimp.

PR #403, opened 2025-07-25
feat: Added issue #273 on news list

Kempu33334:
Added news list.

Note: currently, there I added an option to display news from all past years (as well as just the news items from currently year), but I don't think we need that --- I can delete it if necessary.

_Username: Kempu33334
ID: 2312_


vEnhance:
Cool. I think to keep things simple we can just do news from the current year and delete the other option you mentioned.

Sorry I didn't respond earlier in #273 re the speaker emoji (there is just too much happening this summer in my life), but I'm planning soon to move away from mailchimp (which costs like 30 bucks a month) to having announcements natively managed by OTIS-WEB and using postmark or similar to email them out. I'm also thinking to use postmark for the standard petition / unit check offs, because towards the end of last year some of my emails started getting blocked.

I'll more closely review this PR once that's done (and refactor it so that `get_mailchimp_campaigns` is replaced with whatever new method makes sense), so please give me a bit of time. Appreciate the patience.

In an ideal world, the postmark transition will happen next week right before I have to start reading OTIS applications and letting in the Year XI students. However, my track record this year for keeping self-imposed deadlines is near 0%, so I would not be too optimistic. (Maybe me posting about this deadline here will keep me more honest? Hahaha we'll see.)

vEnhance:
well, i think the point of the PR is to add the other things (opal, markets, etc.) into the news list to go alongside the announcements, since the currently implemented one only lists announcements

vEnhance:
i guess maybe it's a bit redundant now because OPAL/markets/etc. have their own page anyway

maybe let's close this but if we decide we want to enhance the current news view we can start a new PR using some of the code from here

Kempu33334:
> i guess maybe it's a bit redundant now because OPAL/markets/etc. have their own page anyway

Yeah exactly.

> maybe let's close this but if we decide we want to enhance the current news view we can start a new PR using some of the code from here

Ok sounds good.

Issue #384, opened 2025-04-28
Ability to swap units in unit petitions


extouchtriangle:
This is the general reason why I want to suggest this:

It's about to be the summer season and OTIS is about to be closed. I unlock 9 units for the summer. I then realize that I don't really want to do one of them, and I want to swap it for a different unit. Except, when I petition to lock the unit, I have to wait until it gets accepted, and only afterwards can I ask for the new one. I think it would be a lot easier on both sides if there were an option to swap units.

vEnhance:
This was previously suggested as #334 . Anyone who wants to work on this is welcome to (although it may not get merged in time for this year).

Kempu33334:
Close?

PR #405, opened 2025-08-11
feat: Added issue #384 on unit swapping in petitions

Kempu33334:
Added unit swapping in petitions.

Likely to be many things that need to be fixed, so this is basically a first draft.

_Username: Kempu33334
ID: 2363_


Kempu33334:
> From an academic point of view, I'm also still not convinced this is even a net positive because I don't think I should be encouraging people to always be at 9 unlocked units.

That's fair. I don't think this is a good idea anymore...

PR #418, opened 2025-09-20
feat: Added issue #318

Kempu33334:
Added capability to edit problem set submissions from past years.

_Username: Kempu33334
ID: 2312_

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -160,7 +160,9 @@ class PSetQueueList(LoginRequiredMixin, ListView[PSet]):
     template_name = "dashboard/pset_queue_list.html"
 
     def get_queryset(self) -> QuerySet[PSet]:
-        return PSet.objects.filter(status__in=("P", "PA", "PR")).order_by("upload__created_at")
+        return PSet.objects.filter(status__in=("P", "PA", "PR")).order_by(
+            "upload__created_at"
+        )
 
 
 @login_required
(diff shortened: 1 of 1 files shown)


vEnhance:
on whatever device 'you're making these changes in, you should update the evan_django_tools submodule, because it keeps getting a stale pointer in your pull requests

dashboard/views.py
```diff
@@ -263,8 +262,6 @@ def dispatch(
 def resubmit_pset(request: HttpRequest, pk: int) -> HttpResponse:
     pset = get_object_or_404(PSet, pk=pk)
     student = pset.student
-    if student.semester.active is False:
```
vEnhance, line 266: can we verify the logged in user is active for the _current_ semester of OTIS instead? properly speaking random alums should not be able to resubmit their old psets for me to review

dashboard/views.py
```diff
@@ -162,7 +162,6 @@ class PSetQueueList(LoginRequiredMixin, ListView[PSet]):
     def get_queryset(self) -> QuerySet[PSet]:
         return PSet.objects.filter(
             status__in=("P", "PA", "PR"),
```
vEnhance, line 164: drop the comma after this line in that case so the line gets collapsed imo

aincrad/views.py
```diff
@@ -282,7 +280,6 @@ def venueq_handler(action: str, data: JSONData) -> JsonResponse:
         n = 0
         for inquiry in UnitInquiry.objects.filter(
             status="INQ_NEW",
-            student__semester__active=True,
```
vEnhance, line 285: this is in the inquiry logic, so it should not be changed

aincrad/views.py
```diff
@@ -140,7 +139,6 @@ class JSONData(TypedDict):
 
 INQUIRY_VENUEQ_INIT_QUERYSET = UnitInquiry.objects.filter(
     status="INQ_NEW",
-    student__semester__active=True,
```
vEnhance, line 143: this is in the inquiry (petition) logic, so it should not be changed

vEnhance (changes requested):

Issue #301, opened 2023-09-04
Mark Guest Units in Catalogue


Incompleteusern:
Perhaps if this stays a trend.

vEnhance:
seems reasonable if the number of guest units goes above like 3-5 or so

Incompleteusern:
Would it be a good idea to show authors for the unit? It doesn't seem do-able easily (ArrayField only exists for postgress database, so you'd have to make a one to many field to like some `Author` model, etc.)

PR #419, opened 2025-09-23
feat: Added issue #301 on guest units in catalogue

Kempu33334:
Added feature for guest units display in catalogue.

_Username: Kempu33334
ID: 2363_


vEnhance:
I'm going to say we can close the issue as wontfix until there are a larger number of guest units to warrant having an extra field on the model.

PR #449, opened 2026-02-17
feat(rpg): reveal found diamond list to diamond owner (fixes #307)

### LINKED ISSUE #307: Reveal Found Diamond List to Diamond Owner?
Maybe.

pi625:
- Show 'found list' link on diamond solution page to the creator
- Allow creator to access the FoundList view (previously staff-only)
(finished both)

Student ID: 3064

### DIFF
--- a/rpg/tests.py
+++ b/rpg/tests.py
@@ -496,3 +496,34 @@ def test_found_list_creator_access(otis):
     otis.login(charlie.user)
     otis.get_40x("found-listing", a1.pk)
     otis.get_40x("found-listing", a2.pk)
+
+
+@pytest.mark.django_db
+def test_found_list_edge_cases(otis):
+    """Test edge cases for diamond creator access."""
+    alice = StudentFactory.create()
+
+    # Diamond with no creator
+    a1 = AchievementFactory.create(creator=None)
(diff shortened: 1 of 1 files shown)


jatloe:
uh oh

vEnhance (commented):
There should be some unit tests to verify this does the correct thing.

I think some comment about why this would be desirable would also be helpful for context.

pi625:
I added a unit test that passed and fixed dispatch logic to properly check authentication before bypassing staff check and commented a lot more. For the feature itself, I don't see it as significantly desirable or undesirable, just a quality of life type thing(and I wanted to practice coding in a way that wouldn't affect anything). I added a main and edge case unit test and the code works with all other tests. happy to adjust further if needed!

pi625:
yes, i used claude to help write the tests. 
for hijacking dispatch, instead i'll try to create a separate view
in general though, is there any point working on this feature?
(i was away from github for a couple weeks,srry)

vEnhance:
i'm getting feedback that this feature might not be desirable because it would lead to people fighting about trading user diamonds (sigh...), so i'm going to close this for now

PR #458, opened 2026-03-01
feat: add opt-in setting to disable ARCH hints

menezesd:
Add UserProfile.disable_hints boolean field that allows users to hide all hints from the problem archive for pedagogical purposes.

OTIS username: dean

### DIFF
--- a/arch/templates/arch/hint_list.html
+++ b/arch/templates/arch/hint_list.html
@@ -35,24 +35,22 @@
       or <a href="mailto:evan@evanchen.cc">email Evan (evan@evanchen.cc)</a>
       for more personalized help!
     </p>
+  {% elif hints_disabled %}
+    <p>
+      <strong>Hints are disabled in your settings.</strong>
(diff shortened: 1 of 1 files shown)


vEnhance (changes requested):
I think there should be some indication on the hint template itself when hints are disabled, because right now it will just show "there are no hints" for everything (because the blocking is done at the queryset level), and that's a lie --- what actually happened is the user disabled hints, not (necessarily) that there are no hints at all.

there should also be some simple unit test verifying this functionality

vEnhance:
the python part is better, still need to address the previous comments, to repeat:

- I think there should be some indication on the hint template itself when hints are disabled, because right now it will just show "there are no hints" for everything (because the blocking is done at the queryset level), and that's a lie --- what actually happened is the user disabled hints, not (necessarily) that there are no hints at all.
- unit tests

menezesd:
OK I made the change you requested.

PR #478, opened 2026-04-06
feat: inline/attachment settings

norzmaxxer:
I don't know how to pass the settings values into the PDF opener.

2840


norzmaxxer:
Oops, I didn't see that comment.

vEnhance:
can you make it pass CI

vEnhance:
if you delete the migration script the whole project won't work (and it also isn't the formatting issue anyway)

vEnhance:
BTW, I'm sorry I'm being a bit deliberately unhelpful, in that I could probably just run the formatter myself and make it pass CI. However, the fact you didn't do this means that you never got a clone of the repository working to begin with, which means that the code has never been run, and that's the real issue here. Generally when you contribute code to an upstream project you should have at least run it yourself to make sure it works.

(Properly speaking maintainers will ask for unit tests most of the time, but the part you're touching happens to be the one that touches the file system so I'm a bit more willing to overlook it.)

vEnhance:
also, i think the code would be cleaner as

```python
get_from_google_storage(exam.pdfname, user)
```

so you don't have to keep repeating the line `profile, _ = UserProfile.objects.get_or_create(user=request.)` every time you call the function. (As a general rule of thumb if you're a beginner you will make better design decisions if you never allow yourself to copy-paste code.)

norzmaxxer:
The checker says
```
  otisweb/templates/socialaccount/base.html: Broken symlink

vEnhance:
> YAY I HATE UV

nah, uv is great. maybe what you mean to say that you hate pre-commit checks (like ruff, prettier, etc.), however i think that's just skiss

>   otisweb/templates/socialaccount/base.html: Broken symlink

yeah, because you overwrite the symlink with random content (look at the `git diff`). you should undo all the changes to that file since i don't think you even meant to touch it.

norzmaxxer:
I don't know how it got changed. It was modified after running `make fmt`.

norzmaxxer:
Great! Now watch my code crash in prod.

vEnhance:
there are a lot of assert's added (i guess for type safety) so in fact i'm just going to refactor the whole thing to take the entire `request` and do the assertion within the util function

vEnhance:
oh oops i didn't mean to merge my unpushed main stuff in. aiya. ok whatever it's fine.

PR #482, opened 2026-05-05
tests: md5 for faster hashing during tests

vEnhance:
speed up the unit test suite


vEnhance:
this doesn't actually apply here i think because we use `force_login` and don't set a password but might as well have this on by default anyway

PR #484, opened 2026-05-18
docs: update readme installation instructions

cjquines:
add missing command, add `uv run` for the fixture so it uses the uv env

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 help:
 	@echo "Available commands:"
 	@echo "  make install          - Install dependencies with uv"
+	@echo "  make createsuperuser  - Create a superuser"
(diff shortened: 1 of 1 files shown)


vEnhance:
maybe properly speaking what i should do is update all the shell scripts to have `uv run` in them, so they're guaranteed to do the right thing

cjquines:
oh that's probably morally correct, will fix

vEnhance:
i'm about to push an update for the shell scripts rn

vEnhance:
done in https://github.com/vEnhance/otis-web/commit/4ec17032e3b686c12a5c2c66dddf541e003fe361
so we just need the `make createsuperuser` part in here

thanks for catching both of those btw

vEnhance:
고마워

PR #490, opened 2026-06-07
feat: start on in-house OIME system

### LINKED ISSUE #487: Build our own OIME problem system
We'd like to replace the `tube` system (which used to just be invite codes to an external platform) with an in-house one that we control and can link to OTIS student data.
This has two main components, one for authors and one for testsolving.

Everything here must be gated to Verified group (and unit tests to check this).

## Proposal model

Implement a model OIMEProposal with the fields:

- Author (Django user)
- Problem statement
- Answer (integer)
- Solution
- Subject: one of Algebra, Combinatorics, Geometry, Number Theory
- Difficulty, which can be five tiers from 1-5.
- Creation and update date via `auto_add_now`, `auto_add`

Then implement a CreateView and UpdateView.
Users can update their own problems, and staff can update any problems.

## Solving view

Let everyone pick a role upfront: casual or serious.
This might need a new model to track whether someone is casual or serious.

For casual testers:

- Access to all problems and solutions, no locks or timers
- It's fine if they just browse and pick what grabs their interest
- We're not looking for strict contest conditions - their solve times will be unofficial

Serious solvers:

- Get a timer of 5n+5 minutes for a problem with difficulty rating n.
- Once they open the problem, we track the solve times and number of wrong answers and use that to pick statistics

## Comments

There should be a simple comment thread for each problem where the author,
staff, and people who have solved or given up on the problem can share thoughts.
We can implement our own model for this.

vEnhance:
closes #487


tubes/views.py
```diff
@@ -35,17 +50,485 @@ def tube_join(request: HttpRequest, pk: int) -> HttpResponse:
...
+
+
+# ---------------------------------------------------------------------------
+# OIME: Setup / onboarding
+# ---------------------------------------------------------------------------
+
+
+@verified_required
+def oime_setup(request: HttpRequest) -> HttpResponse:
+    """First-time setup: create OIMEContributor and optionally enrol in active year."""
```
vEnhance:
```suggestion
    """First-time setup: create OIMEContributor and optionally enrol in active year."""
``` enrol -> enroll

PR #495, opened 2026-06-28
feat: show problem statement source on arch problem page

isobreaked:
Added "Show problem source" collapsible element below the problem statement to show the copyable source code of a problem in a box.


vEnhance:
thanks, let's try it.

(the PR's get squashed into a single commit, so it doesn't actually matter how many commits you have on a PR)

PR #502, opened 2026-08-02
feat(exams): extreme laziness

vEnhance:
it's legit faster to ask claude to write a script to move the due dates forward two years than to actually change the dates with importexport like we used to every year

plus while claude is thinking i can get up and walk around


vEnhance:
> it's legit faster to ask claude to write a script to move the due dates forward two years than to actually change the dates with importexport like we used to every year

just kidding i definitely got scammed

not that the ci failure is actually related to this PR but still smh

PR #515, opened 2026-08-05
feat: make ARCH problem source executable

isobreaked:
cleaned arch problem source feature to actually be executable latex. removed span and p tags, replaced ol and ul with enumerate and itemize environments.


vEnhance:
... uh.

something must be broken in the pipeline because the output really should be the orginal LaTeX as evil chin outputs it and not this `<p>` nonsense anyhow

let me see what's going on

vEnhance:
i think this is obsoleted by #518 now. I think #495 wasn't probably implemented and I just never noticed because I never opened the source dropdown. Still, a bit embarrassing for me that it took so long for me to realize this.

PR #516, opened 2026-08-07
fix(dashboard): shrunken emoji on locked units when zoomed in Firefox

vEnhance:
Locked units were dimmed with `filter: brightness(60%) saturate(60%)` on the whole `<tr>`. Firefox renders a filtered element through an intermediate surface, and color emoji drawn into that surface come out far too small once the page is zoomed, so the icons on locked rows shrank while unlocked rows stayed correct.

Drop the row filter and set the dimmed subject backgrounds directly. The new values are the existing subject colors run through the same brightness/saturate transform in sRGB, so the rendered rows are unchanged (verified pixel-for-pixel against the old rules, including Bootstrap's striped overlay). The locked text color and the Secret-subject colors are likewise pre-dimmed, and the thumbnail filter picks up the brightness the row used to contribute.

Locked-row emoji are now drawn at full color rather than dimmed, since a filter is what dimmed them; the rows stay clearly muted via background, text color, and italics.

### DIFF
--- a/otisweb/static/css/otis.css
+++ b/otisweb/static/css/otis.css
@@ -121,64 +121,81 @@ table.table tr.subjectH:nth-child(even) td {
   background-color: #ecb3ef;
 }
 
-/* Locked units are dimmed.  The colors below are the subject colors above run
-   through `brightness(60%) saturate(60%)`, precomputed rather than applied as a
-   `filter:` on the row.  Firefox draws a filtered element through an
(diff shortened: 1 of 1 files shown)


vEnhance:
ok now that i'm actually looking into this myself i can't repro the original bug
going to hold off working on this unless i can get feedback from users the bug actually exists

vEnhance:
i haven't gotten any indication from anyone that this can be reproduced, so i'm closing it now

PR #522, opened 2026-08-08
feat(arch): 404 if solution not found

cjquines:
Somewhat RFC. Right now we 500 if arch finds the statement but not the solution; this changes it to a 404.

Semantically these _should_ be 404s. In the arch-sol case it's expected that we can't find things in the bucket sometimes, so a 500 is inappropriate. Also, they fire both a critical _and_ an error log, and my SRE brain believes that the steady state of prod should have no 5xy errors.

### DIFF
--- a/arch/tests.py
+++ b/arch/tests.py
@@ -241,6 +241,31 @@ def test_view_solution(otis):
     assert resp.status_code == 404
 
 
+@pytest.mark.django_db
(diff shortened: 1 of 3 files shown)


vEnhance:
I've actually been meaning to do this, so thanks for reading my mind :)

PR #528, opened 2026-08-15
perf: optimize scroll wheel blocker

vEnhance:
dude my head is such a mess


cjquines:
in general i recommend not fighting the browser but. shrug

vEnhance:
yeah idk maybe we can revert if more issues come up

this particular feature request has come up several times this year already though. idk whose idea it was to make scroll behave like that

PR #532, opened 2026-08-16
tests: never again smh

vEnhance:
... so this is not what i asked claude for but i guess in principle more tests can't hurt?

well okay let me just stick this in and if the tests start causing issues later i'll just revert it


vEnhance:
ok i have no idea what test_view_guards.py is doing ugh

vEnhance:
whatever

PR #534, opened 2026-08-17
feat(tubes): mark proposal author's comments with a badge

vEnhance:
Comments on the OIME proposal detail page written by the contributor who wrote the problem now show a small "Author" badge next to the commenter name, so readers can spot the author's replies at a glance.


vEnhance:
3 lines of code is worth 30 lines of test apparently :face_with_spiral_eyes:

cjquines:
i often delete claude tests sometimes because they're just like. too much

vEnhance:
> i often delete claude tests sometimes because they're just like. too much

i should maybe start doing that

i feel like i always felt guilty for not writing enough tests, but that advice might have been for the pre-AI world.

PR #541, opened 2026-08-20
fix(payments): make Stripe credits idempotent and handle reversals

vEnhance:
Bug fix (billing correctness), from the same security review as #539. Covers two related findings in the Stripe webhook.

**Duplicate deliveries credited an invoice repeatedly.** Every valid `checkout.session.completed` delivery ran `invoice.total_paid += amount` and inserted a `PaymentLog`, with no check that the payment intent had been seen before. Stripe retries webhooks as a matter of course, so an ordinary retry could make a student's debt look paid twice over.

`process_payment()` now runs in a transaction and returns the existing log untouched if that payment intent was already credited. Manually entered payments carry no payment intent and are still always credited. A partial `UniqueConstraint` on `PaymentLog.stripe_id` (positive amounts, non-empty id) backs this up, but SQLite honours partial indexes and MySQL does not, so the transaction also takes `select_for_update()` on the invoice — that, not the constraint, is what stops two simultaneous deliveries in production. The credit itself moved to an `F()` expression so concurrent payments to one invoice can't clobber each other's read-modify-write.

**Refunds and disputes never gave the credit back.** The webhook acknowledged every verified event but acted only on `checkout.session.completed`, so a refunded or charged-back payment left the invoice showing as paid until someone fixed it by hand.

`charge.refunded`, `charge.dispute.created`, and `charge.dispute.funds_withdrawn` are now handled by `process_reversal()`, which records the reversal as a separate `PaymentLog` with a negative amount rather than editing the original, keeping the history auditable. It reverses only the portion not already taken back, so redelivered events, the two dispute events for one dispute, and a series of partial refunds all settle at the right total. Once a payment is fully reversed, it and its reversal rows are flagged `refunded` together — the aincrad sync at `aincrad/views.py:505` recomputes `total_paid` by summing un-refunded logs, and flagging only the original would have double-subtracted.

An event whose payment intent has no matching credit, and a reversal event with no payment intent at all, are logged and ignored rather than crashing the webhook.

Also adds `cents_to_usd()` for the conversion that was inline, and help text on `refunded` saying what it means for the totals.

Migration `0038` adds the constraint and the help text.

Tests in `payments/tests.py` cover: a duplicated completion event crediting once; refund and dispute events reversing the credit; a redelivered refund and a dispute reported twice both being no-ops; partial refunds accumulating correctly; and `total_paid` staying equal to the sum of un-refunded logs, which is the invariant the aincrad sync depends on.

Verified with `make check` (pyright clean, no missing migrations) and `make test`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01675RPzfCV5n9HTLU5bkkEQ)_


vEnhance:
imagine having a math grad student write your payment system though

PR #545, opened 2026-08-20
fix: require POST for job claims, inquiry cancels, and Mystery unlocks

vEnhance:

Bug fix — closes finding [13] of the static security report (CWE-352, CSRF).

Django's CSRF middleware exempts safe methods, so a view that mutates on `GET` can be triggered by a cross-site top-level navigation carrying the victim's session cookie. Three views mutated before any method check:

| View | Route | Effect on GET |
| --- | --- | --- |
| `payments.views.job_claim` | `job/claim/<pk>/` | assigns a job to the victim |
| `roster.views.cancel_inquiry` | `inquiry/cancel/<pk>/` | cancels the victim's unit petition |
| `roster.views.unlock_rest_of_mystery` | `mystery-unlock/{easier,harder}/` | swaps the victim's Mystery unit |

### `job_claim` and `cancel_inquiry`

Both are already submitted as POST forms by `payments/job_detail.html` and `roster/inquiry.html`, and their existing tests already POST, so they just gain `@require_POST`. It sits below `@login_required` so an anonymous visitor still gets a login redirect rather than a bare 405.

### `unlock_rest_of_mystery`

This one isn't linked from any template — the URL is handed out for students to visit directly, so making it POST-only would break the workflow it exists to serve. Instead a `GET` renders a new confirmation page whose form carries a CSRF token, and only the `POST` performs the swap. The permission check for actually having Mystery unlocked still runs first, so a student without it gets the same error as before rather than a confirmation page that would fail.

The page deliberately doesn't name the unit that's about to be unlocked; the view's `get_object_or_404` for it stays after the method check so that a confirmation page can't be used to peek at what's behind Mystery.

The legacy `mystery_unlock/` (underscore) paths still 302 to the hyphenated ones; a redirected POST becomes a GET, so those now land on the confirmation page, which is the intended destination anyway.

### Tests

- `test_claim_rejects_get` and `test_cancel_inquiry_rejects_get` — assert 405 and that the object is unchanged.
- `test_mystery_unlock_get_only_confirms` — a GET renders the confirmation page and leaves the curriculum alone.
- The two existing Mystery tests now POST.

`make check` is clean (0 pyright errors, no missing migrations) and `roster` + `payments` pass — 60 tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01675RPzfCV5n9HTLU5bkkEQ)_

### DIFF
--- a/payments/tests.py
+++ b/payments/tests.py
@@ -408,6 +408,20 @@ def test_claim_limits(otis) -> None:
 
 
 @pytest.mark.django_db
+def test_claim_rejects_get(otis) -> None:
+    verified_group = GroupFactory(name="Verified")
+    alice: User = UserFactory.create(username="alice", groups=(verified_group,))
+    otis.login(alice)
+    otis.post_ok("worker-update", data={"notes": "hi"}, follow=True)
+    job = JobFactory.create()
+
+    # a cross-site navigation is a GET, so claiming can't be reachable that way
+    assert otis.get("job-claim", job.pk).status_code == 405
+    job.refresh_from_db()
+    assert job.assignee is None
+
+
+@pytest.mark.django_db
 def test_job_update_permissions(otis) -> None:
     verified_group = GroupFactory(name="Verified")
     alice = WorkerFactory.create(user__username="alice", user__groups=(verified_group,))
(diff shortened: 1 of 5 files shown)


vEnhance:
man i love vibe coding from my phone while waiting to board an aircraft

i hope prod stays up for the duration of my flight

cjquines:
maybe i'll set up an on-call rotation 😭

vEnhance:
you make it sound like we have an engineering team

PR #546, opened 2026-08-21
perf(core): add etags to files for caching

cjquines:
watermarking is expensive for long PDFs, so avoid doing so if client's copy isn't stale, by using an etag (and add cache-control as well, while we're at it i guess). this mildly annoyed me lol

this does mean that the timestamp represents the last time the file was downloaded, not the last time it was accessed, but i think that's fine

### DIFF
--- a/core/tests.py
+++ b/core/tests.py
@@ -1,8 +1,11 @@
 import io
 import json
 from typing import Any
+from unittest import mock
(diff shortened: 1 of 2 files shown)


vEnhance:
thanks! much appreciated.

PR #550, opened 2026-08-22
refactor: pretty much all the JS

vEnhance:
javascript gives me the heebie-jeebies man


cjquines:
you can use ts if you add a build step! and that might give you some more peace of mind

vEnhance:
there's overall not much javascript in otis-web (intentionally, because i'm so much more familiar with python), so i don't know if i feel that having a build step is worth it yet. if we ever start building more intensive javascript systems, then i agree moving to typescript is the right call. i have noticed that now that i have claude, i'm a lot more willing to throw in a javascript file to do things like #526 or #549, so it's conceivable the amount of javascript might increase pretty rapidly.

it would require moving out all the inline scripts though. (right now, if a JS snippet is <100ish lines and only used in one place, i just put it directly in the template.)

PR #552, opened 2026-08-23
feat(aincrad): bulk-sync OPAL puzzle PDFs from a local directory

vEnhance:
rather than rclone


cjquines:
now i have to write a puzzle called tetrogram

vEnhance:
lmao

https://2025.teammatehunt.com/puzzles/tetrogram for the audience

PR #553, opened 2026-08-23
fix(otisweb): disallow empty option

cjquines:
regression caused by #550 . the empty option thing caused several rendering issues, so we just disallow it. adds a bunch of placeholders as well for consistency

### DIFF
--- a/arch/forms.py
+++ b/arch/forms.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 from django import forms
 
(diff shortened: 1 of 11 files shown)


cjquines:
i forget if gh notifies on pr ready for review, but pr ready for review

vEnhance:
> i forget if gh notifies on pr ready for review, but pr ready for review

i don't get emails for that, so a comment like that is useful so i know to look

PR #554, opened 2026-08-23
feat(tubes): make problem table sortable

cjquines:
- add tubes fixtures
- make sortable tables three-state (asc, desc, default)
- make tubes table sortable


cjquines:
(unclear if we actually want tubes table to be sortable always, or just for casual mode)

vEnhance:
i think it's fine to just have it always sortable

PR #559, opened 2026-08-25
fix(tubes): anchor timer to wall-clock

cjquines:


vEnhance:
고마워 <3

vEnhance:
that explains whoever was saying they stopped the timer by switching tabs

PR #567, opened 2026-08-30
feat(tubes): casual browser header polish and voting from the card

vEnhance:
Follow-up to #566, from prod feedback on the casual per-subject browser.

## Describe what kind of changes you made here

A mix of a layout fix, an enhancement, and one small behaviour change.

**Layout fix — the card header badges.** They carried `px-2 py-2`, which made
them taller than the plain title text beside them; combined with
`align-items-start` the two sat on different baselines. The row is now
`align-items-center`, the badges use the default badge padding, and the gap
between them is `gap-1` so they take less width next to a long title. The
difficulty badge also moved ahead of the heart, matching the order on the
problem's own page.

**Enhancement — the difficulty badge filters.** Clicking one filters the
browser to that difficulty, or clears the filter if it is the one already
applied. While a filter is on the badges render solid primary, the same
treatment the toolbar buttons use. The per-row target is built in the view,
so these are plain links and nothing changes about how the filtering works.

**Enhancement — voting from the card.** The heart showed vote state but did
nothing, while on the problem's own page the heart *is* the vote button. It
is one here too now, so triaging a subject can be read-statement, heart,
scroll without leaving the page. A vote returns to the page it was cast from,
including sort, difficulty and page number; the target is rebuilt server-side
from a posted subject code plus re-parsed settings rather than echoed back, so
a forged form can only ever point at a subject page of this browser, and
anything unrecognized falls back to the problem page where the detail-page
button still lands.

**Behaviour change — what upvoting requires.** `can_upvote` was defined as
`can_see_solution`, one flag answering both "may I read the solution" and "may
I rate this". That left a ranked contributor unable to vote during a running
fight, with the statement open in front of them, while a casual browser four
lines up could vote having read only the statement and never the solution. It
is now gated on having seen the *statement*: in ranked mode from the moment a
fight starts rather than only once it finishes. Voting on a problem you have
never opened stays blocked, which is the case worth preventing, and archived
problems stay frozen.

Worth flagging that this is only observable through the casual browser —
`proposal_detail` redirects a ranked contributor mid-fight to the fight page
and one with no fight to the start screen, and neither renders a heart. The
browse cards were the only place the old rule bit. Landing copy updated to
match.

## Testing

`make fmt`, `make check` (pyright clean) and 127 tubes tests pass. New
coverage for the badge filter toggle, the vote round-trip and its return
target, forged and malformed return params, voting during a running fight,
and voting before a statement has been seen.

---
_Generated by [Claude Code](https://claude.ai/code/session_014zqjcRQjq1aiSwYjcBEcZ1)_

### DIFF
--- a/otisweb/static/css/otis.css
+++ b/otisweb/static/css/otis.css
@@ -404,14 +404,6 @@ img.emoji {
 .oime-upvote-btn:active {
   transform: scale(0.95);
 }
-.oime-upvote-btn:disabled {
-  cursor: default;
-  opacity: 0.65;
-}
-.oime-upvote-btn:disabled:hover {
-  transform: none;
-  box-shadow: none;
-}
 
 /* OIME proposal table collapse toggle */
 .oime-table-toggle::before {
--- a/otisweb/static/css/otis.css
+++ b/otisweb/static/css/otis.css
@@ -404,6 +404,14 @@ img.emoji {
 .oime-upvote-btn:active {
   transform: scale(0.95);
 }
+.oime-upvote-btn:disabled {
+  cursor: default;
+  opacity: 0.65;
+}
+.oime-upvote-btn:disabled:hover {
+  transform: none;
+  box-shadow: none;
+}
 
 /* OIME proposal table collapse toggle */
 .oime-table-toggle::before {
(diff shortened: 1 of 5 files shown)


vEnhance:
really we should just be redoing the whole view altogether i feel like the current one is super clunky

PR #568, opened 2026-08-30
feat(tubes): redo OIME Tubes layout

vEnhance:
more views make happy


vEnhance:
this feels so much better cleaner to me now

let's hope people agree ㅋㅋㅋㅋ

(if not they should just keep using the table view and ignore the new ones)

PR #576, opened 2026-09-04
feat(opal): explain "call in" on a close answer

vEnhance:
Explains what "keep going" means specifically for `CALL IN X`.

I understand why GPH internally bans this now lmao

### DIFF
--- a/opal/views.py
+++ b/opal/views.py
@@ -563,12 +563,14 @@ def show_puzzle(
                         reverse("opal-finish", args=(puzzle.hunt.slug, puzzle.slug))
                     )
             elif attempt.is_close:
-                messages.warning(request, f"Keep going for {puzzle.title}...")
+                # codespell:ignore-next-line callin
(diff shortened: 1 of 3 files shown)


opal/views.py
```diff
@@ -564,6 +564,11 @@ def show_puzzle(
                     )
             elif attempt.is_close:
                 messages.warning(request, f"Keep going for {puzzle.title}...")
+                if answerize(attempt.guess).startswith("CALLIN"):
+                    messages.info(
+                        request,
+                        'In puzzle hunts, "call in X" is an instruction to submit the answer X',
+                    )
```
cjquines, line 571: in gph we have a generic "this answer → this message" thing. i'm wondering if it's worth just doing that rather than special casing

vEnhance: i'm on the fence about this. maybe i'll merge this as a band-aid for now and then file an issue for a generic message system.

PR #586, opened 2026-09-11
refactor(dashboard): redo components / dismiss-news

vEnhance:
i think i'm going to cherry pick a18fefa out maybe


vEnhance:
yea okay cherry picked perf(core) out since it's not related to this, and we'll let claude do the cleanup

PR #588, opened 2026-09-13
feat(core): add timestamp tag

cjquines:
adds `<time title="...">` tags around all timestamps

we have around a two-to-one split of `DATETIME_FORMAT` (which i think is "Sept. 12, 2026, midnight") to ISO, and it doesn't look like there's a consistent way we choose one over the other. now would be a good time to think about a policy here, like:

- switch everything to ISO
- switch everything to `DATETIME_FORMAT`
- use ISO for admin-y shaped things and tables with lots of data, use `DATETIME_FORMAT` otherwise

i also dislike passing `"DATE_FORMAT"` right now, and am considering making it an enum with `date`, `iso-date`, `relative`, etc. i'll change this once we decide on the datetime policy


cjquines:
i'm also not sure how i feel about how this adds `{% load otis-extras %}` to a bunch of stuff

vEnhance:
i don't mind loading otis extras in more places

vEnhance:
> use ISO for admin-y shaped things and tables with lots of data, use DATETIME_FORMAT otherwise

i think this is what i was doing in my head, although i absolutely don't think i was actually consistent with it

"switch everything to ISO" is definitely simplest and most compact though

vEnhance:
> use ISO for admin-y shaped things and tables with lots of data, use DATETIME_FORMAT otherwise

yeah i think my vote is to do it this way.

i agree with using an enum, so it seems like we just need something like these five:

- relative
- date
- isodate
- time
- isotime

cjquines:
well, there's one spot where we use `"M d"` (`Sep 12`) as added in https://github.com/vEnhance/otis-web/commit/ff24a0bff809c7289d2d41607fb2961b5089ac93 ; i'll just make that an `isodate`. i think i'll also make the argument required so the spelled out vs. iso is made explicit

cjquines:
alright, ready for review i think. i let claude made some judgment calls about where to use spelled out vs. iso; i spot-checked and it looks fine

vEnhance:
cool thanks much let's ship this. (just added one code comment)

PR #589, opened 2026-09-13
polish(dashboard): 2 decimal pts for spades

ddhliu89:
The number of decimal places on the spades meter has been changed from 1 to 2 because markets award spades to an accuracy of 2 decimal places.

-
my otis-web id is 3287
This is my first time doing all of this so if I do something stupidly wrong please point it out to me.


vEnhance:
looks fine

vEnhance:
should be live on prod within a few minutes

Issue #118, opened 2022-11-19
Anon survey feedback


vEnhance:
We'd like to be able to reply to anonymous student feedback.

vEnhance:
Need a mechanism like the following:

- Student generates key K not sent to server
- Server stores comments in plaintext together with a hash H(U,K) for U = uid
  and K = this key
- Evan can reply to feedback and whatever
- If the student wants to edit or follow up, they present K to the server to
  prove they were original author of survey feedback.

The problem is how to store K. The student could generate a password, but they
might forget it. Could store client side cookie, but that's per device.

Best solution is to see if there's a way to integrate with Django built-in auth
somehow, but I need to research to see if or how that's possible.

Incompleteusern:
> If the student wants to edit or follow up, they present K to the server to
prove they were original author of survey feedback.

The server can correlate K to an ip address which the server can later correlate to a student, I think

vEnhance:
I don't know if I trust IP addresses to be that static either.

AmoleR:
> > If the student wants to edit or follow up, they present K to the server to
> > prove they were original author of survey feedback.
> 
> The server can correlate K to an ip address which the server can later correlate to a student, I think

IP addresses is the worst way to go about it. You reset your router, your MAC changes, and so does your IP address. Or you go somewhere else. Whatever.

> Best solution is to see if there's a way to integrate with Django built-in auth
> somehow, but I need to research to see if or how that's possible.
Isn't a great idea, but can you make `K` the hash of the student's ID with a random string? Like obviously not 100% protection, as the student IDs are clearly enumerable, but it's a lot of work to do it.

I don't know how else you want to do it. Curious: what feature do u think Django auth would have that solves this?

Incompleteusern:
I feel like that if this is done in Django, wouldn't the auth token sent with the request reveal the student.

jondoglover:
Can the server generate the key, automatically email it to the student, and immediately delete the key? It never stores the key or the email that way. If the student wants to follow up, then they can find the key that was emailed to them, and the server can just check the inputted key against the hash to verify.

vEnhance:
Right now the server does not have any email capability at all.

Incompleteusern:
- Force every student to generate an anonymous PGP key. 
- Survey is paired with a PGP key when entering feedback. 
- Evan responds to feedback in a public channel open to all.
[...]

PR #592, opened 2026-09-15
feat(surveys): add survey models

cjquines:
Design decisions here:

- Survey fields are hard-coded because they haven't changed in forever
- Prevent duplicate survey responses
- GM and instructor feedback are separate tables, so they can be independently signed or anonymous
- GMs screen instructor feedback before forwarding it
- Anonymous responses are entirely unlinked from students (discussion below)

(Claude wrote all the actual code, of course.)

If we really wanted to do anon replies (as per #118), then my thinking is:

- In my threat model, if someone has access to request logs they can deanonymize no matter what, due to things like timing, or physical DB access, or whatever. So there's no advantage to public-key cryptography over something symmetric, and I'm not worrying about request leaks.
- Suppose someone had access to the DB. For them to be unable to deanonymize a response, a student must have some sort of private-knowledge key, stored externally from the DB. Both localStorage or cookies are browser-specific and can get wiped, so to guarantee anon reply receipt, a student must store the key externally, which they can always fail to do. This also roughly doubles the number of views.

If we're willing to bite the bullet of implementation complexity, plus students possibly forgetting their key, then we can implement anon replies. I'm not sure if it's worth it, since I don't have as much context behind why we want anon replies.


vEnhance:
I agree that it's not worth the additional complexity of having an external key.

This mostly looks good to me. I think the one thing I might change is that the random primary key would be better on the `SurveyCompletion` than on `GMFeedback` and `InstructorComment`. I think for `GMFeedback` it is a little useful to me for them to just be ascending PK's in the default way since having submission-time order helps me orient myself in responding to them.

I also don't think I need to screen instructor feedback, and we can just deliver that directly.

vEnhance (changes requested):
see above

vEnhance (approved):
great, thanks much for the speedy work

PR #596, opened 2026-09-17
feat(surveys): add survey form

cjquines:
this pr
- adds the backend for submitting surveys, and corresponding tests
- adds the frontend for filling out surveys, and showing response for survey
- **(!)** changes achievement granting to be a bulk admin action, for more anonymity reasons

remaining work
- add views for looking at survey responses, marking as read, replying to them
- add view for viewing reply to survey responses
- make frontend for the main survey view actually look good
- add the survey emoji on the grid

<details>
<summary>screenshots</summary>

image

image

image

</details>


vEnhance:
> (!) changes achievement granting to be a bulk admin action, for more anonymity reasons

... sigh. You're right that if we timestamp the achievement unlock, then it technically defeats the anonymity. I guess that's OK that we do it as a bulk create when the survey closes.

vEnhance:
I wonder whether we should make it clearer that if you select an anonymous reply, then it will be impossible to view the feedback later on?

... actually here's an idea (that I should have brought up at the model step). What if, when someone submits a survey, we generate a random UUID for that survey response as well, and have that be a token that someone can look at (if they either want to see their previous response, or even if Evan replies to it?). If someone loses the UUID, then so bad so sad, but that's fine. On the other hand, a lot of people will have the URL's autosaved in the browser history, and we can include messaging to make it clear that they must keep the URL if they ever want to see their survey response again.

I have a second request for model change (sorry) --- maybe just a Boolean that says "read by Evan" or something. I think for a lot of feedback I won't have text to right, but want to acknowledge I've seen the feedback. So maybe we can just have a small Boolean for that to make things a bit more systematic rather than me writing the word "Thanks" over and over.

vEnhance:
sorry I just remember we had an `is_read` already

vEnhance:
so i think the point is to make that user-facing and also make it and explicit GM button or something (i.e. is_read isn't set to True unless I take a non-passive action, just opening the feedback shouldn't mark it as read)

cjquines:
> ... actually here's an idea (that I should have brought up at the model step). What if, when someone submits a survey, we generate a random UUID for that survey response as well, and have that be a token that someone can look at (if they either want to see their previous response, or even if Evan replies to it?). If someone loses the UUID, then so bad so sad, but that's fine. On the other hand, a lot of people will have the URL's autosaved in the browser history, and we can include messaging to make it clear that they must keep the URL if they ever want to see their survey response again.

i think that's fair. i'll follow-this on another pr

cjquines:
> so i think the point is to make that user-facing and also make it and explicit GM button or something (i.e. is_read isn't set to True unless I take a non-passive action, just opening the feedback shouldn't mark it as read)

yeah i think that's the plan for when we add the gm-facing survey response view

vEnhance:
> i think that's fair. i'll follow-this on another pr

kk should i merge this one now then?

cjquines:
yeah sure

PR #597, opened 2026-09-17
chore(fixtures): mark fixtures as generated

cjquines:
also seeds faker

makes things like #596 have a smaller-lookin diff


vEnhance:
oh that's cool didn't know about this feature

ty

PR #598, opened 2026-09-18
feat(surveys): add private links to anonymous feedback

cjquines:
question looks like this
image

<details>
<summary>on submit, anon</summary>
image
</details>
<details>
<summary>private link</summary>
image
</details>


surveys/templates/surveys/components/gm_feedback.html
```diff
@@ -0,0 +1,29 @@
...
+  <div class="card mb-3 border-primary" data-testid="survey-reply">
+    <div class="card-header">
+      Reply from Evan
+      {% if gm_feedback.replied_at %}({% timestamp gm_feedback.replied_at "time" %}){% endif %}
+    </div>
+    <div class="card-body">{{ gm_feedback.reply|linebreaks }}</div>
+  </div>
+{% elif gm_feedback.is_read %}
+  <p class="text-muted" data-testid="survey-read">Evan has read your feedback.</p>
+{% endif %}
```
cjquines, line 36: reminder to test this when i add the ui for reading/replying

cjquines:
alright, after this there should be one more PR that'll finish up this feature i think

surveys/templates/surveys/components/gm_feedback.html
vEnhance: i think this would be cuter if it was like a 👍 emoji or something like that as a read indicator, kinda like how comments on serena's blog are ack'ed by a heart reaction ( https://blog.anserena.com/moving-to-nyc/ e.g.)

surveys/templates/surveys/gm_feedback_detail.html
```diff
@@ -0,0 +1,26 @@
...
+{% block title %}
+  {{ survey }}
+{% endblock title %}
+{% block layout-content %}
+  <p>
+    <a href="{% url "survey-list" %}">← Back to surveys</a>
+  </p>
+  <div class="alert alert-success">
+    If you don't have the diamond for this survey,
+    it'll be awarded when the survey closes.
```
vEnhance: maybe we can clarify awarded automatically (not even needing for the end-user to enter a code, which is unusual --- rn the only other diamond that behaves this way is OPAL iirc)

vEnhance (changes requested):
seems good, two minor cosmetic requests and then we can merge this

surveys/templates/surveys/components/gm_feedback.html
```diff
@@ -0,0 +1,36 @@
+{% load otis_extras %}
+<h2>Your feedback</h2>
+<div class="card mb-3">
+  <div class="card-header">Free comments</div>
+  <div class="card-body">
+    {{ gm_feedback.essay|linebreaks }}
+    {% if gm_feedback.is_read and not gm_feedback.reply %}
```
vEnhance: tbh i think we can show this even if `gm_feedback.reply` is true

vEnhance (changes requested):

surveys/templates/surveys/components/gm_feedback.html
cjquines: sure, i think that's cute

PR #599, opened 2026-09-18
feat(surveys): add inbox view 

cjquines:
this should finish up the survey feature

- add view for looking at survey responses, marking as read, replying to them
- make survey response view show evan reply
- make frontend for survey list better
- add survey emoji to grid
- adds portal news when survey opens

<details>
<summary>inbox view</summary>
image
</details>

<details>
<summary>reply from evan</summary>
image
</details>

<details>
<summary>survey list for admin</summary>
image
</details>

survey icon is here between invoice and calendar
image

(and reminder to myself to update parent-info after this lands)


cjquines:
@vEnhance ready for review!

vEnhance:
thanks much, let's run with this

PR #600, opened 2026-09-18
feat(roster)!: new standing field to replace booleans

vEnhance:
rather than three different booleans, compress to a single choicefield with some new features


cjquines:
oh thank gods, these bits always confused me

cjquines:
some relevant reading: 
- https://ntietz.com/blog/that-boolean-should-probably-be-something-else/ argues that bools in db columns should usually be timestamps (`is_confirmed` -> `confirmed_at`) or enums
- https://www.jmduke.com/posts/one-status-field-per-model.html argues that your data should have at most one "status"-like field

vEnhance:
good to know, will keep that in mind

off my head i know that the `Assistant` object definitely violates that and i got bitten by that recently. refactoring that is not a big priority but maybe i should open an issue anyway so i don't forget that

vEnhance:
opened #601, although it's not actionable yet

BTW if there are other parts of the codebase that look bad to you please just open issues with the refactor label, or ping me otherwise

PR #604, opened 2026-09-19
fix: Minor fix to reflect new rules

Kempu33334:
Changed the market description from alpha = 1 to alpha = 2.

ID: 3261


vEnhance:
The default value is changed to alpha=1 in 2026-2027.

